PR #69, opened 2019-01-06
Fix action logger

ishaiborovoy:


ishaiborovoy:
@doroncy please review, I fixed the action logger for the slider, thanks

PR #71, opened 2019-01-07
Adding disabled to slider

ishaiborovoy:


projects/ui-framework/src/lib/slider/slider.component.scss
```diff
@@ -10,6 +10,9 @@
         .mat-slider-track-background {
           background-color: $grey-300;
         }
+        .mat-slider-thumb {
+          transform: scale(0.7);
+        }
```
omrihecht: does this really solve it? how?

ishaiborovoy: It just make the thumb the same size when the slider is enabled.  The thumb will be visible in disabled mode, that what decided today with @doron

omrihecht: ok, you can also colour 50% on the right with the full colour and 50% of the left in the empty colour

PR #75, opened 2019-01-08
Start develop the datepicker

ishaiborovoy:


projects/ui-framework/src/lib/datepicker/datepicker.component.html
```diff
@@ -0,0 +1,5 @@
+<mat-form-field>
+  <input matInput [matDatepicker]="picker" placeholder="Choose a date" (dateChange)="onDateChange($event)">
+  <mat-datepicker-toggle matSuffix [for]="picker"></mat-datepicker-toggle>
+  <mat-datepicker #picker></mat-datepicker>
+</mat-form-field>
```
doroncy: pay attention that all the texts should be come from outside

ishaiborovoy: Yes, I'll change it to @Input.
The input needs also to be replaced with `b-input` which somehow need to know how to work with `[matDatepicker]` like the above:
`<input matInput [matDatepicker] ...`

doroncy: talked about it with omri, it wont be b-input, it will just input and omri is setting some global style for input.

PR #120, opened 2019-01-17
style_fix

omrihecht:
style_fix

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/input/input.component.ts
+++ b/projects/ui-framework/src/lib/form-elements/input/input.component.ts
@@ -1,11 +1,10 @@
-import { Component, EventEmitter, forwardRef, Input, OnInit, Output, ViewChild } from '@angular/core';
-import { InputAutoCompleteOptions, InputEventType, InputTypes } from './input.enum';
+import { AfterViewInit, Component, ElementRef, forwardRef, OnInit, ViewChild } from '@angular/core';
+import { InputEventType } from './input.enum';
 import { inputAttributesPlaceholder } from '../../consts';
 import { MatInput } from '@angular/material';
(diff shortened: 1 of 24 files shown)


projects/ui-framework/src/lib/form-elements/input/input.component.ts
```diff
@@ -98,4 +106,10 @@ export class InputComponent extends BaseFormElement implements OnInit {
       this.propagateChange(value);
     }
   }
+
+  hasTranscludeContent(node: HTMLElement): boolean {
+    return node.childNodes.length > 0
```
ishaiborovoy: I think you can do it just with return, no need if then here

omrihecht: how?

projects/ui-framework/src/lib/form-elements/input/input.component.ts
```diff
@@ -98,4 +106,10 @@ export class InputComponent extends BaseFormElement implements OnInit {
...
+      ? node.childNodes.length > 1 || !includes(this.emptyNodeNames, node.childNodes[0].nodeName)
+      : false;
+  }
 }
```
ishaiborovoy: return node.childNodes.length > 1 && (node.childNodes.length > 1 || !includes(this.emptyNodeNames, node.childNodes[0].nodeName));

Not sure what is more readble ... So let you decide...

projects/ui-framework/src/lib/form-elements/input/input.component.ts
ishaiborovoy: return node.childNodes.length > 0 && (node.childNodes.length > 1 || !includes(this.emptyNodeNames, node.childNodes[0].nodeName));

Not sure what is more readble ... So let you decide...

PR #221, opened 2019-02-18
Fix ngOnChange

ishaiborovoy:


projects/ui-framework/src/lib/form-elements/lists/multi-select/multi-select.component.ts
```diff
@@ -66,7 +66,9 @@ export class MultiSelectComponent extends BaseSelectPanelElement implements OnCh
   }
 
   ngOnChanges(changes: SimpleChanges): void {
-    this.triggerValue = this.getTriggerValue(this.value);
+    if (changes.value) {
+      this.triggerValue = this.getTriggerValue(this.value);
```
omrihecht: it's cleaner for sure, but i will be surprised if this solves it, did you try other lifecycle hooks?

ishaiborovoy: I solved the endless loop in form-generator and how to set the value, see the pool request there in hibob-web-client

ishaiborovoy: I solved the endless loop in form-generator and how to set the value, see the pool request there in hibob-web-client

ishaiborovoy: This is just to ensure you update when value changed and not on every other change

PR #243, opened 2019-02-25
Adding some events to the grid

ishaiborovoy:


projects/ui-framework/src/lib/table/table/table.component.html
```diff
@@ -1,5 +1,6 @@
 <ag-grid-angular
   #agGrid
+  style="width: 100%; height: 100%"
```
omrihecht: ? is this just for development?

omrihecht: you can move it to stories instead, not important atm

ishaiborovoy: It does not define the size but the point that it should be 100%, and that what let virtual scroll to work.  see https://www.ag-grid.com/javascript-grid-width-and-height/

PR #259, opened 2019-02-28
Add DateRage picker with story

eliraneliassy:


omrihecht:
will be done in another pr

PR #299, opened 2019-03-12
side menu 3 - exposing the component, interface and the mocks

benbaler:


benbaler:
@omrihecht i want to expose the mock as well

projects/ui-framework/src/public_api.ts
```diff
@@ -84,3 +84,7 @@ export {
 // Menu
 export { MenuModule } from './lib/overlay/menu/menu.module';
 export { MenuItem } from './lib/overlay/menu/menu.interface';
+// Side Menu
+export { SideMenuModule } from './lib/navigation/side-menu/side-menu.module';
+export { SideMenuOption, OptionComponent } from './lib/navigation/side-menu/side-menu-option/side-menu-option.interface';
+export { getSideMenuOptionsMock } from './lib/navigation/side-menu/side-menu.mock';
```
omrihecht: why would you want to expose the mock? it should not be part of the library, only expose modules/interfaces/enums, things consumed.

benbaler: for testing, but ok. i will remove

PR #303, opened 2019-03-13
side menu 4

benbaler:
1. add the add_doc icon
2. handle empty options array

- [x] tests

### DIFF
--- a/projects/ui-framework/src/lib/navigation/side-menu/side-menu.component.ts
+++ b/projects/ui-framework/src/lib/navigation/side-menu/side-menu.component.ts
@@ -1,6 +1,7 @@
 import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
 import { SideMenuOption } from './side-menu-option/side-menu-option.interface';
 import head from 'lodash/head';
+import size from 'lodash/size';
 
 @Component({
   selector: 'b-side-menu',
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/navigation/side-menu/side-menu.component.ts
```diff
@@ -17,7 +18,9 @@ export class SideMenuComponent implements OnInit {
   }
 
   ngOnInit() {
-    this.onSelectOption(head(this.options).id);
+    if (size(this.options) > 0) {
```
omrihecht: it's not important but don't be afraid to use `this.options.length`
also consider moving to OnChanges, it will allow adding options async and updating them

benbaler: sure? i thought we have to use lodash all the time for everything

omrihecht: it's safer, it will not fall if options is undefined.
also i think maybe it's better to move this to OnChanges, to make it async and consumer will be able to update the list if they need

benbaler: ok i will do it in next PR, i need this to be merged for revi to continue work on it

PR #314, opened 2019-03-17
Pavel/back button

galloween:
Added disabled state prop.

### DIFF
--- a/projects/ui-framework/src/lib/buttons-indicators/buttons/buttons.scss
+++ b/projects/ui-framework/src/lib/buttons-indicators/buttons/buttons.scss
@@ -59,5 +59,9 @@
     border: 1px solid $grey-300;
     background-color: $grey-200;
     color: $grey-500;
+
+    .mat-icon {
+      opacity: 0.2;
+    }
   }
 }
--- a/projects/ui-framework/src/lib/buttons-indicators/buttons/buttons.scss
+++ b/projects/ui-framework/src/lib/buttons-indicators/buttons/buttons.scss
@@ -59,9 +59,5 @@
     border: 1px solid $grey-300;
     background-color: $grey-200;
     color: $grey-500;
-
-    .mat-icon {
-      opacity: 0.2;
-    }
   }
 }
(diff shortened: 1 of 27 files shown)


projects/ui-framework/src/lib/style/colors.scss
```diff
@@ -1,11 +1,11 @@
...
+$grey-100: #F8F7F7; // rgb(248,247,247)
+$grey-200: #F3F2F2; // rgb(243,242,242)
+$grey-300: #E9E9E9; // rgb(233,233,233)
+$grey-500: #D2D2D2; // rgb(210,210,210)
+$grey-600: #9D9D9D; // rgb(157,157,157)
+$grey-700: #535353; // rgb(83,83,83)
+$grey-800: #303030; // rgb(48,48,48)
 
-$boxShadowColor: rgba(83, 83, 83, 0.2);
+$boxShadowColor: rgba($grey-700, 0.2);
```
omrihecht: cool, didn't know you can do that

projects/ui-framework/src/lib/buttons-indicators/buttons/buttons.scss
```diff
@@ -59,5 +59,9 @@
     border: 1px solid $grey-300;
     background-color: $grey-200;
     color: $grey-500;
+
+    .mat-icon {
+      opacity: 0.2;
```
omrihecht: what's the opacity for?

projects/ui-framework/src/lib/style/colors.scss
galloween: yes you can!
I also like to keep colors in rgb (or at least comment hex color as rgb - which i did in the colors file), because then you can read it and see which color is darker, greener, etc. sass will compile to hex in the end anyway, but rgb seems more readable...

projects/ui-framework/src/lib/buttons-indicators/buttons/buttons.scss
galloween: when disabled the icon should get lighter. now i understand, i should have changed the fill on svg to new color instead.

PR #326, opened 2019-03-19
Pavel/cards tests

galloween:


projects/ui-framework/src/lib/cards/card-add/card-add.component.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  describe('component', () => {
+    it('should set input Title inside <b-display-3> element', () => {
+      component.title = 'hello';
+      fixture.detectChanges();
+      const bDisplay3Element = fixture.debugElement.query(
+        By.css('b-display-3')
+      );
+      expect(bDisplay3Element.nativeElement.innerText).toEqual('hello');
+    });
+    it('should set input Subtitle inside <p card-bottom> element', () => {
```
omrihecht: why inside p tag?

omrihecht: we should use one of the typography tags i think

projects/ui-framework/src/lib/cards/card/card.component.scss
```diff
@@ -28,8 +29,9 @@ b-menu {
   position: absolute;
   top: 0;
   right: 0;
-  opacity: 0;
   transition: opacity 0.2s;
+  opacity: 0;
+  display: none;
```
omrihecht: if you switch from display none to block, the transition will not work i think. we can use *ngIf and angular transition for this, or just show hide without opacity transition, it might be meaningless

projects/ui-framework/src/lib/cards/card/card.component.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      fixture.nativeElement.style.width = '200px';
+      component.text =
+        'Compensation update with a very long text that cuts off after 4 lines of text. And here is another very long text that should not be displayed at all.';
+      fixture.detectChanges();
+      const bDisplay3Element = fixture.debugElement.query(
+        By.css('b-display-3')
+      );
+      expect(bDisplay3Element.nativeElement.scrollHeight).toBeGreaterThan(
+        bDisplay3Element.nativeElement.clientHeight
+      );
```
omrihecht: nice

projects/ui-framework/src/lib/cards/card/card.component.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+    it('should create menu element when menu configuration is passed', () => {
+      component.menu = [{ label: '' }];
+      fixture.detectChanges();
+      const menuElement = fixture.debugElement.query(By.css('b-menu'));
+      expect(menuElement).toBeTruthy();
+    });
+    it('should NOT create menu element when invalid menu configuration is passed', () => {
```
omrihecht: you can also add test that it's not being created for undefined

projects/ui-framework/src/lib/cards/card-add/card-add.component.spec.ts
galloween: it needs to be like body text. so can be any non-special element.

projects/ui-framework/src/lib/cards/card/card.component.scss
galloween: yes you are right. I put display none to avoid the menu being focused/clicked etc. i will put pointer events none/all instead.

projects/ui-framework/src/lib/cards/card/card.component.spec.ts
galloween: yes, there was that test (we did it with you), I removed it, it think it might be excessive (?)

projects/ui-framework/src/lib/cards/card/card.component.scss
omrihecht: alternatively we can use ngIf with transition, this will also work

projects/ui-framework/src/lib/cards/card/card.component.spec.ts
omrihecht: I think there is no harm, it checks 2 different things

PR #330, opened 2019-03-21
Pavel/chips

galloween:
Renamed Badge to Chip, now uses mat-chip internally (at the moment does not add any additional fucntionality)


projects/ui-framework/src/lib/buttons-indicators/chips/chip/chip.component.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      .compileComponents()
+      .then(() => {
+        fixture = TestBed.createComponent(ChipComponent);
+        component = fixture.componentInstance;
+      });
+  }));
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
omrihecht: we can remove this test

projects/ui-framework/src/lib/buttons-indicators/chips/chip/chip.component.ts
```diff
@@ -0,0 +1,24 @@
+import { Component, Input, HostBinding } from '@angular/core';
+import { ChipType } from '../chips.enum';
+
+@Component({
+  selector: 'b-chip, [b-chip]',
+  template: `
+    <mat-basic-chip [selectable]="false" [disableRipple]="true">
+      <ng-content></ng-content>
```
omrihecht: what's the ng-content for?

galloween: it allows to be used like this <b-chip>hello</b-chip> or <p b-chip>item</p>
material chip also behaves like that (you can pass [value] binding, or <mat-chip>text</mat-chip>)
also maybe we need to pass avatar or icon or button there in the future.
just for flexibility, basically. i can remove it, if we want stricter api

projects/ui-framework/src/lib/buttons-indicators/chips/chip/chip.component.spec.ts
galloween: yes, it was in button test I used as reference. but its unnecessary of course. any other test will fail if component is not created.

PR #332, opened 2019-03-24
Pavel/collapsible

galloween:
First draft of Collapsible panel component (no responsive or mobile views)


projects/ui-framework/src/lib/style/style.scss
```diff
@@ -1,8 +1,5 @@
 @import './variables.scss';
 @import './typography.scss';
-@import './colors.scss';
```
omrihecht: why are these removed?

omrihecht: the mixins in these files are exposed through this import to other projects, is it somewhere else now?

galloween: only colors is removed - its being imported by variables now, because I put the css vars colors to scss variables

galloween: oh i did not know the styles file was imported by other projects. i will put back the shapes then, along with the mixins and functions files I created, I guess

PR #336, opened 2019-03-25
Pavel/new mini preloader

galloween:
+ new mini preloader 
+ implementation for Collapsible
+ css color vars also stored as scss vars


omrihecht:
nice idea, but we need to develop it a bit more

PR #342, opened 2019-03-28
Pavel/card table

galloween:
basic version
- no component transclusion
- no sorting
- no tests


galloween:
sorry I see it doesn't have latest commits, ignore for now

PR #343, opened 2019-03-28
Pavel/card table

galloween:
basic card table
- no component transclusion
- no sorting
- no tests


projects/ui-framework/src/lib/layout/card-table/card-table/card-table.component.ts
```diff
@@ -0,0 +1,27 @@
...
+  styleUrls: ['./card-table.component.scss']
+})
+export class CardTableComponent {
+  constructor() {}
+
+  @Input() table: CardTableData;
+
+  @HostBinding('attr.role') string = 'table';
+
+  getHeaderId(index: number): string {
```
omrihecht: what is this used for?

projects/ui-framework/src/lib/layout/card-table/table-card/table-card.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  isString(val: any): boolean {
+    return typeof val === 'string';
+  }
+
+  isArray(obj: any): boolean {
+    return Array.isArray(obj);
+  }
+
+  getHeaderId(index: number): string {
```
omrihecht: looks similar, should be in a service that both used if nessessary, but it looks like this is something that should be part of the interface, let's talk about it

projects/ui-framework/src/lib/style/style.scss
```diff
@@ -6,15 +6,16 @@
 @import '~ag-grid-community/dist/styles/ag-grid.css';
 @import 'quill.snow';
 
-
 body {
   margin: 0;
   padding: 0;
   color: $grey-700;
-  @include BodyFont(normal, normal, 12px);
+  @include BodyFont(normal, normal, $font-size-body);
```
omrihecht: cool, let's go over the storybook once we are done with the table and check if there are other places with this.
should be in a new pr

projects/ui-framework/src/lib/style/typography.scss
```diff
@@ -23,7 +24,7 @@ $font-size-body: 12px;
   text-transform: uppercase;
 }
 
-@mixin BodyFont($style, $weight, $size, $lineHeight: 1.5) {
+@mixin BodyFont($style, $weight, $size, $lineHeight: $line-height-body) {
```
omrihecht: these are good changes, but let's separate them in prs and have a pr dedicated to these kind of changes, it will help keep prs smaller

projects/ui-framework/src/lib/layout/card-table/card-table/card-table.component.ts
galloween: for semantics, accessibility and machine readability mostly. its part of HTML5 ARIA standards, like labeledby etc. we're not using native table element, but we can mark it as one with roles. I also set roles for rows and cells etc...

projects/ui-framework/src/lib/layout/card-table/table-card/table-card.component.ts
galloween: its already abstracted to a external util function in the other PR.
but I got a new idea about this table, I will not need the id's at all eventually.

projects/ui-framework/src/lib/style/style.scss
galloween: its a small change - I just put the 12px in a variable, so that I can use the variable in other places. it shouldnt affect anything

projects/ui-framework/src/lib/style/typography.scss
galloween: same as above, i will put it in separate PR, ok

omrihecht:
moved to https://github.com/hibobio/bob-style/pull/348

PR #345, opened 2019-03-28
Pavel/card table insert component

galloween:
service to pass component as cell data (with ngContent)
*** merge after PR #343


projects/ui-framework/src/lib/layout/card-table/card-table-utils.ts
```diff
@@ -0,0 +1,12 @@
+
```
omrihecht: you can remove empty lines

galloween: i dont have them locally, actually. maybe the file got commited before the autoformat did its thing...

galloween:
yes this PR should come after the previous one

omrihecht:
moved to https://github.com/hibobio/bob-style/pull/348

PR #347, opened 2019-03-31
table col def

omrihecht:
need to add tests to the table component


projects/ui-framework/src/lib/table/table/table.component.spec.ts
```diff
@@ -33,111 +100,85 @@ describe('TableComponent', () => {
...
-        },
-      ]);
-      tick(1);
-      expect(component.sortChanged.emit).toHaveBeenCalledWith({
-        colId: 'fullName',
-        sort: 'asc'
-      });
-    }));
-  });
+  // describe('Table events', () => {
```
ishaiborovoy: Why leave the comments here?

omrihecht: i need to get back to the tests, but i want to do the lists first

PR #352, opened 2019-04-01
Pavel/card table width calc

galloween:
the widths algorithm is a bit on the complex side, 
but i think it may be useful later if we have similar case where we need to set widths - then we will abstract it to service...


projects/ui-framework/src/lib/layout/card-table/card-table/card-table.component.ts
```diff
@@ -1,27 +1,89 @@
...
+    }
+
+    if (totalWidth() < 100) {
+      cellWidths = cellWidths.map(
+        width => width + (100 - totalWidth()) / cellWidths.length
+      );
+    }
+
+    return cellWidths;
+  }
```
omrihecht: let's move this out to a service, will also be easier to test this

galloween:
moved here: https://github.com/hibobio/bob-style/pull/361

PR #354, opened 2019-04-01
fix dialog

benbaler:
fix dialog, allow to resolve the action with false to leave the dialog open

### DIFF
--- a/projects/ui-framework/src/lib/overlay/dialog/dialog.component.spec.ts
+++ b/projects/ui-framework/src/lib/overlay/dialog/dialog.component.spec.ts
@@ -151,6 +151,24 @@ describe('DialogComponent', () => {
         tick();
         expect(spyMatDialogRef.close).toHaveBeenCalled();
       }));
+    it('should leave dialog open if action resolves false', () => {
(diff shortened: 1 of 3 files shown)


projects/ui-framework/src/lib/overlay/dialog/dialog.component.spec.ts
```diff
@@ -151,6 +151,18 @@ describe('DialogComponent', () => {
         tick();
         expect(spyMatDialogRef.close).toHaveBeenCalled();
       }));
+    it('should leave dialog open if action resolves false', () => {
+      fakeAsync(() => {
```
omrihecht: this test checks that things don't exist, they don't exist by default so how do you know the test works?
for example if i only test `expect(spyMatDialogRef.close).not.toHaveBeenCalled();`
the test will pass

benbaler: ok

benbaler: done

PR #361, opened 2019-04-02
Pavel/card table comp factory service

galloween:
latest version with all the changes


projects/ui-framework/src/lib/layout/card-table/card-table.module.ts
```diff
@@ -1,19 +1,21 @@
...
+    TableCardCellComponent,
     TableCardComponent,
-    CardTableComponent
+    CardTableComponent,
+    ComponentRendererComponent
   ],
   imports: [CommonModule],
-  exports: [TableCardCardTableCellComponent, TableCardComponent, CardTableComponent],
-  providers: [ComponentFactoryService]
+  exports: [TableCardCellComponent, TableCardComponent, CardTableComponent],
```
galloween: no need to export TableCardCellComponent and TableCardComponent

PR #367, opened 2019-04-07
Pavel/tooltip directive

galloween:
component & directive


projects/ui-framework/src/lib/services/component-renderer/component-renderer.component.ts
```diff
@@ -136,7 +137,7 @@ export class ComponentRendererComponent implements OnInit, OnDestroy {
 
   private insertComponent(comp: RenderedComponent): void {
     this.reset();
-    this.componentRef = this.createComponent(comp, true);
+    this.componentRef = this.createComponent(comp, true, this.container);
```
omrihecht: this could use a test

projects/ui-framework/src/lib/services/truncate-tooltip/truncate-tooltip.component.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  ngOnDestroy(): void {
+    this.resizeSubscription.unsubscribe();
+  }
+
+  private applyTextElementStyle(): void {
+    const computedStyle = getComputedStyle(this.textElement),
+      fontSize = parseFloat(computedStyle.fontSize),
+      lineHeight =
```
omrihecht: not sure i understand this bit, can you explain what `applyTextElementStyle` does?

galloween: calculates maximum height from font-size & line-height, then sets the appropriate css-truncation/line-clamp rules

projects/ui-framework/src/lib/services/component-renderer/component-renderer.component.ts
galloween: rendered-component has tests.
here nothing really changes, createComponent method is now gets passed the container, instead of always using this.container.
i was thinking of abstracting that method to a service, to use in tooltip, but then gave up on that idea.

for the rendered-component component itself, nothing really changed in the way that it works

PR #410, opened 2019-04-17
b-link component

tomeraronovsky:
b-link component


projects/ui-framework/src/lib/buttons-indicators/link/link.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="b-link">
```
omrihecht: no need for wrapper

projects/ui-framework/src/lib/buttons-indicators/link/link.component.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../style/colors';
+@import '../../style/typography';
+
+:host {
+  .b-link {
```
omrihecht: remove wrapper

projects/ui-framework/src/lib/buttons-indicators/link/link.component.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    component.config = { url: 'https://app.hibob.com', text: 'Learn more' };
+    component.color = LinkColor.primary;
+    component.openInNewWindow = true;
+    fixture.detectChanges();
+  });
+
+  it('link should be with the right url and text', () => {
+    const aElement = fixture.debugElement.query(By.css('a'));
+    expect(aElement.componentInstance.config.text).toEqual('Learn more');
+    expect(aElement.componentInstance.config.url).toEqual('https://app.hibob.com');
```
omrihecht: cool

projects/ui-framework/src/lib/buttons-indicators/link/link.component.ts
```diff
@@ -0,0 +1,16 @@
...
+import { LinkColor } from './link.enum';
+
+@Component({
+  selector: 'b-link',
+  templateUrl: './link.component.html',
+  styleUrls: ['./link.component.scss']
+})
+export class LinkComponent {
+  @Input() config: Link;
+  @Input() openInNewWindow: boolean;
```
omrihecht: give default value

projects/ui-framework/src/lib/buttons-indicators/link/link.enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum LinkColor {
+  none = 'dark',
```
omrihecht: none = 'none', no need for dark color, should be body color, also in scss

projects/ui-framework/src/lib/buttons-indicators/link/link.component.ts
tomeraronovsky: its false by default

projects/ui-framework/src/lib/buttons-indicators/link/link.stories.ts
```diff
@@ -0,0 +1,55 @@
...
+const note = `
+  ## Slider Element
+  #### Module
+  *SliderModule*
+
+  #### Properties
+  Name | Type | Description | Default value
+  --- | --- | --- | ---
+  config | object | link configuration - text, url |
+  openInNewWindow | boolean | open link in new window or not  | false (optional)
```
omrihecht: optional is on the param, not the default value
also need to implement the default in the controller

projects/ui-framework/src/public_api.ts
```diff
@@ -67,6 +67,9 @@ export { ChipType } from './lib/buttons-indicators/chips/chips.enum';
 export { InfoStripModule } from './lib/buttons-indicators/info-strip/info-strip.module';
 export { InfoStripComponent } from './lib/buttons-indicators/info-strip/info-strip.component';
 export { Link } from './lib/buttons-indicators/info-strip/info-strip.types';
+// Link Module
+export { LinkModule } from './lib/buttons-indicators/link/link.module';
+export { LinkComponent } from './lib/buttons-indicators/link/link.component';
```
omrihecht: export interface as well

projects/ui-framework/src/lib/buttons-indicators/link/link.component.scss
```diff
@@ -0,0 +1,16 @@
+@import '../../style/colors';
+@import '../../style/typography';
+
+:host {
+  a {
+    text-decoration-line: none;
+    line-height: 14px;
+    color: $grey-700;
```
omrihecht: does color: initial work?

projects/ui-framework/src/lib/buttons-indicators/link/link.stories.ts
```diff
@@ -0,0 +1,55 @@
...
+import { values } from 'lodash';
+const color = values(LinkColor);
+
+const linkStories = storiesOf(ComponentGroupType.ButtonsAndIndicators, module).addDecorator(
+  withKnobs
+);
+const template = `<b-link
+  [config]="config"
+  [openInNewWindow]="openInNewWindow"
+  [color]="color">
```
galloween: I think either all config should be in the config input, or everything in separate inputs.
not logical when url is in config object and the _blank in separate input...
@omrihecht

projects/ui-framework/src/lib/buttons-indicators/link/link.component.scss
galloween: no. 'initial' will not work (will give default browser link color).
but 'unset' will (will give text color of parent).

projects/ui-framework/src/lib/buttons-indicators/link/link.stories.ts
tomeraronovsky: I had interface that contain url & text so I wanted to use it here. @omrihecht WDYT ?

galloween: maybe the new interface should be extended with additional/optional properties?...
(then its backwards compatible)

export interface Link {
  text: string;
  url: string;
  openInNewWindow?: boolean;
  color?: string;
}

omrihecht: I agree, missed that, there should be one interface for the link @tomeraronovsky please update

PR #412, opened 2019-04-17
implement b-link component in b-info-strip

tomeraronovsky:
implement b-link component in b-info-strip + css fix from designers

### DIFF
--- a/projects/ui-framework/src/lib/buttons-indicators/info-strip/info-strip.component.scss
+++ b/projects/ui-framework/src/lib/buttons-indicators/info-strip/info-strip.component.scss
@@ -16,12 +16,8 @@
 
     .content {
       text-align: left;
-      .url {
-        a {
-          color: $grey-700;
-          font-weight: 700;
-          text-decoration-line: none;
-        }
+      .text {
+        font-weight: 300;
       }
     }
   }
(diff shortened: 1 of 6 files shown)


projects/ui-framework/src/lib/buttons-indicators/link/link.component.scss
```diff
@@ -6,6 +6,7 @@
     text-decoration-line: none;
     line-height: 14px;
     color: $grey-700;
```
omrihecht: see @galloween comment from yesterday about the color (inherit or unset), also let's remove line-height, should be the default line-height from the body (it's 1.5 by default)

projects/ui-framework/src/lib/buttons-indicators/info-strip/info-strip.component.scss
```diff
@@ -16,12 +16,8 @@
...
     .content {
       text-align: left;
-      .url {
-        a {
-          color: $grey-700;
-          font-weight: 700;
-          text-decoration-line: none;
-        }
+      .text {
+        font-weight: 300;
```
omrihecht: i don't think we have font-weight 300, do you need it? it probably makes no difference in the ui

projects/ui-framework/src/lib/buttons-indicators/link/link.component.scss
```diff
@@ -4,8 +4,8 @@
 :host {
   a {
     text-decoration-line: none;
-    line-height: 14px;
-    color: $grey-700;
+    color: inherit;
+    font-weight: 500;
```
omrihecht: 👍

PR #416, opened 2019-04-18
add new 3 icons

danazelniker:


projects/ui-framework/src/lib/icons/icons.enum.ts
```diff
@@ -54,5 +54,7 @@ export enum Icons {
   employee_ref = 'employee_ref',
   docs_empty = 'docs_empty',
   baseline_info_icon = 'baseline_info_icon',
-  doc_icon = 'doc_icon',
+  approve_badge = 'approve_badge',
+  arrow_forward = 'arrow_forward',
+  pending_badge = 'pending_badge'
```
benbaler: the icons svgs are not aligned, hope it will look nice when you will use it.

Issue #43, opened 2018-12-25
overlay panel


doroncy:
need to investigate the CDK overlay, this will be used in panels and dropdowns

omrihecht:
[overlay panel](https://hibobio.github.io/bob-style/?selectedKind=Overlay&selectedStory=Panel&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fnotes%2Fpanel)
cc: @Rotemfischbob @levben

omrihecht:
- [ ] add position strategy
- [ ] add basic position strategy input

galloween:
done.

PR #418, opened 2019-04-18
panel update

omrihecht:
- update default position strategy to open above center by default
 - add input option to open above or below
 - add open on hover behaviour and inputsel

connected to:
https://github.com/hibobio/bob-style/issues/43

### DIFF
--- a/projects/ui-framework/src/lib/overlay/panel/panel.component.ts
+++ b/projects/ui-framework/src/lib/overlay/panel/panel.component.ts
@@ -1,20 +1,18 @@
-import { Component, Input, OnDestroy, TemplateRef, ViewChild, ViewContainerRef } from '@angular/core';
+import { Component, Input, OnDestroy, OnInit, TemplateRef, ViewChild, ViewContainerRef } from '@angular/core';
 import { CdkOverlayOrigin, FlexibleConnectedPositionStrategy, Overlay, OverlayConfig, OverlayRef } from '@angular/cdk/overlay';
 import { TemplatePortal } from '@angular/cdk/portal';
-import invoke from 'lodash/invoke';
-import { PanelPositionService } from './panel-position.service';
+import { PanelPositionService } from './panel-position-service/panel-position.service';
 import { Subscription } from 'rxjs';
-import { PanelSize } from './panel.enum';
-import concat from 'lodash/concat';
-import compact from 'lodash/compact';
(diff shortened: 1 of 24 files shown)


projects/ui-framework/src/lib/overlay/panel/panel.component.ts
```diff
@@ -23,25 +21,49 @@ export class PanelComponent implements OnDestroy {
   @Input() panelClass: string;
   @Input() panelSize = PanelSize.medium;
   @Input() showBackdrop = true;
+  @Input() panelDefaultPosVer = PanelDefaultPosVer.above;
```
galloween: i think inputs can be simplified to 'size', 'class' etc, because we know we're talking about panel.
unless its a breaking change - then nevermind ;)

projects/ui-framework/src/lib/overlay/panel/panel.stories.ts
```diff
@@ -8,18 +8,21 @@ import { TypographyModule } from '../../typography/typography.module';
...
+import { PanelDefaultPosVer, PanelSize } from './panel.enum';
 import { values } from 'lodash';
 
 const buttonStories = storiesOf(ComponentGroupType.Overlay, module).addDecorator(withKnobs);
 
 const panelSize = values(PanelSize);
+const panelDefaultPosVer = values(PanelDefaultPosVer);
 
 const template = `
 <b-panel style="position: absolute; top: 20px; left: 20px;"
```
galloween: i believe the storybook styling should be moved to storyTemplate (wrapping <div style="....)

projects/ui-framework/src/lib/style/style.scss
```diff
@@ -11,6 +11,7 @@ body {
   padding: 0;
   color: $grey-700;
   @include BodyFont;
+  -webkit-font-smoothing: antialiased;
```
galloween: ha! its my find 🥇 ;) delivered through Dana. i have some more of these coming in next PRs.
we are not using a reset/reboot strategy in bob-style, i have a proposal for it!

projects/ui-framework/src/lib/overlay/panel/panel.stories.ts
omrihecht: true, i plan to do this, also add a bit better examples there

PR #421, opened 2019-04-18
menu disabled and tests

omrihecht:
add disabled to menu trigger
add tests to menu component

### DIFF
--- a/projects/ui-framework/src/lib/navigation/menu/menu.component.html
+++ b/projects/ui-framework/src/lib/navigation/menu/menu.component.html
@@ -1,6 +1,9 @@
-<div [matMenuTriggerFor]="childMenu" (menuOpened)="onOpenMenu()">
+<content [matMenuTriggerFor]="childMenu"
+         (menuOpened)="onOpenMenu()"
+         [ngClass]="{'disabled': disabled}"
+         class="menu-trigger-wrapper">
   <ng-content select="[menu-trigger]"></ng-content>
-</div>
+</content>
 
 <mat-menu #childMenu="matMenu"
           (closed)="onCloseMenu()"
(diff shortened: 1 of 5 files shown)


projects/ui-framework/src/lib/navigation/menu/menu.component.html
```diff
@@ -1,14 +1,17 @@
...
   <ng-content select="[menu-trigger]"></ng-content>
-</div>
+</content>
 
 <mat-menu #childMenu="matMenu"
           (closed)="onCloseMenu()"
           [xPosition]="menuDir">
   <span *ngFor="let child of menu">
     <!-- Handle branch node menu menu -->
     <span *ngIf="child.children && child.children.length > 0">
```
galloween: you can replace all/any <span(or other tag) *ngIf> with <ng-container *ngIf> to save on DOM elements

projects/ui-framework/src/lib/navigation/menu/menu.component.scss
```diff
@@ -5,6 +5,12 @@
 
 :host {
   display: inline-block;
+  .menu-trigger-wrapper {
```
galloween: :host .menu...
will compile to
[ng123] .menu..[ng123] .  (3 classes specificity)

my humble advise is:
only wrap in :host something that is :ng-deep
otherwise it just adds unnecessary specificity.
when inside angular component you dont even have to give something a class. you can use element selector: p, div, span, etc - because it will be scoped/compiled to p[ng123], div[ng123] etc...
the attribute specificity equals class specificity, so angular already does it for you (gives you a class).
i believe css/scss in components should be written flat. (:host, span, div, p.., no nesting)
then its much easier to override.
you will never have to do 
".ql-snow.ql-container .ql-formats .ql-picker-options .ql-picker-item ... blahblah"
sass nesting is cool, but only for global css!

PR #423, opened 2019-04-21
Pavel/rich text editor

galloween:
I managed to get the rich-text to send changes on blur or on change (depending on settings in formControl)
	to get it working:
- base-form-element missed implementation of registerOnChange (registers function to call on blur)
- i am injecting ngControl to read the options from there (updateOn chnage/blur)


projects/ui-framework/src/lib/form-elements/base-form-element.ts
```diff
@@ -11,17 +10,18 @@ export abstract class BaseFormElement implements ControlValueAccessor {
...
-  propagateChange: Function = (_: any) => {
-  };
+  propagateChange: Function = (_: any) => {};
 
   registerOnChange(fn: any): void {
     this.propagateChange = fn;
   }
 
   registerOnTouched(fn: any): void {
+    this.onTouched = fn;
```
omrihecht: what does this do?
@ishaiborovoy please note, this affects all form elements

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.spec.ts
```diff
@@ -1,47 +1,90 @@
...
+})
+class TestComponent {
+  constructor() {}
+  rteControl = new FormControl();
+}
 
-describe('TableComponent', () => {
-  let component: RichTextEditorComponent;
-  let fixture: ComponentFixture<RichTextEditorComponent>;
+xdescribe('RichTextEditorComponent', () => {
```
omrihecht: WIP?

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.ts
```diff
@@ -89,6 +98,7 @@ export class RichTextEditorComponent extends BaseFormElement
   ];
 
   @Input() type?: RTEType = RTEType.primary;
+  @Input() sendChangeOn = RTEchangeEvent.blur;
```
omrihecht: why does it have to be one or the other? what if you want both?

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.ts
```diff
@@ -209,6 +228,11 @@ export class RichTextEditorComponent extends BaseFormElement
 
     this.editor.root.addEventListener('blur', () => {
       this.blurred.emit(this.latestOutputValue);
+      if (!this.writingValue && this.sendChangeOn === RTEchangeEvent.blur) {
```
omrihecht: what is this for?
`!this.writingValue`

projects/ui-framework/src/lib/form-elements/base-form-element.ts
ishaiborovoy: Since it just adding functionality seems it's ok

galloween: its similar to registerOnChange - it registers an external callback provided by formControl as local method, to be called on blur

registerOnChange registers local propagateChange method
registerOnTouched registers local onTouched method

then in RTE I manually call onTouched to propagate changes on blur

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.spec.ts
galloween: yes and its also in another branch, so ignore the tests in this PR

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.ts
galloween: that's how formControl expects it - its either/or
I have local blurred/changed events as backup as well, that fire anyway not depending on this setting.

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.ts
galloween: there is a writeValue method that formControl uses to init RTE with a value from the model. At that time formControl doesnt want the value to be propagated to avoid a loop.
so when its called Im setting writingValue to true. and on user-originating change im setting it back to false which allows to propagate again.

projects/ui-framework/src/lib/form-elements/base-form-element.ts
galloween: registerOnTouched was a missing implementation of ControlValueAccessor interface. so it actually should be there for all form elements.

omrihecht: cool

projects/ui-framework/src/lib/form-elements/rich-text-editor/rich-text-editor.component.ts
omrihecht: ok, maybe indicate flag in the naming,
basically you only want false on the init right?

galloween: formControl may write any time actually. on init is most probable, but not limited to on init.
the method is called writeValue thats why i called it writingValue.
what do you think is a more appropriate name?
isWritingValue ?

omrihecht: it's a boolean, the name is ok, ignore the comment

PR #473, opened 2019-05-01
side menu bold option on select

benbaler:


benbaler:
@omrihecht i don't know how to make the icon bold on select, i pass it as component from outside.
any ideas?

projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.scss
```diff
@@ -16,6 +16,12 @@
       }
     }
 
+    &.selected {
+      .option-display-name {
+        font-weight: bold;
```
danazelniker: we don't want to add it in bob style?

danazelniker:
@omrihecht My bad... it's in bob-style... too many branches I guess 🤦‍♀️

projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.scss
omrihecht: why was this merged? we said we are going to do this in bob-style

PR #483, opened 2019-05-02
Template text editor | RTE component

Yossitrx:
## Status 
**READY**

## Description
Placeholder is function ready! 
In this PR I'm connecting the panel and single select to the RTE.
Plus output and input are working. 

## Todos
- [x] Tests

BOB-STYLE

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte-form-element.abstract.ts
+++ b/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte-form-element.abstract.ts
@@ -1,21 +1,22 @@
 import {
-  ElementRef,
-  Input,
-  ViewChild,
-  Output,
-  EventEmitter,
-  OnChanges,
-  SimpleChanges,
   ChangeDetectorRef,
+  ElementRef,
+  EventEmitter,
   Injector,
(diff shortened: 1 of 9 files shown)


projects/ui-framework/src/lib/form-elements/rich-text-editor/rte-form-element.abstract.ts
```diff
@@ -155,6 +159,10 @@ export abstract class RTEformElement extends BaseFormElement
     if (this.value !== val || !this.latestOutputValue) {
       this.value = val || '';
       if (this.editor) {
```
galloween: this check (if (this.editor)) is only for this.editor.setContents.
so should be something like:

this.value = val && this.controls.includes(RTEControls.placeholders) ? this.placeholderRteConverterService.toRte(val, placeholders) : (val || '');

if (this.editor) {
	this.editor.setContents(this.editor.clipboard.convert(this.value));
}

Yossitrx: why should I parse the content and then have the possibility not to set it to the editor?

PR #486, opened 2019-05-05
add list footer

omrihecht:
add list footer component
move clear action to footer.

multiList:
image

multiSelect:
image

---

singleList:
image

singleSelect:
image

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/lists/list.scss
+++ b/projects/ui-framework/src/lib/form-elements/lists/list.scss
@@ -1,6 +1,7 @@
 @import '../../style/typography';
 @import '../../style/colors';
 @import '../../style/shapes';
+@import '../../style/variables';
 
 @mixin listOption {
   @include BodyFont;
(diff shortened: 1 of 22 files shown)


projects/ui-framework/src/lib/form-elements/lists/list.scss
```diff
@@ -63,3 +63,14 @@
     }
   }
 }
+
+@mixin selectPanel {
+  display: block;
+  box-sizing: border-box;
+  border: 1px solid $grey-500;
+  border-top: none;
+  border-bottom-right-radius: 4px;
```
galloween: we have a variable for border-radius now

omrihecht: where?

omrihecht: i saw, maybe we should move it to shapes?

galloween: not sure. if someone will be looking for variable for something, the first (and probably only) place he will check will be the variables file....

PR #491, opened 2019-05-06
move ListFooterAction to interface file

omrihecht:


projects/ui-framework/src/lib/form-elements/lists/list-footer/list-footer.component.ts
```diff
@@ -1,12 +1,7 @@
 import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { ButtonSize, ButtonType } from '../../../buttons-indicators/buttons/buttons.enum';
 import { Icons } from '../../../icons/icons.enum';
-
-export interface ListFooterActions {
-  clear?: true;
-  cancel?: true;
-  apply?: true;
-}
+import { ListFooterActions } from './list-footer.interface';
```
galloween: why not put it in the list.interface.ts file?

projects/ui-framework/src/lib/form-elements/lists/list-footer/list-footer.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface ListFooterActions {
+  clear?: true;
```
galloween: true is not a type...
Am I missing something?

omrihecht: what do you mean by type?
it's an interface

omrihecht: oh you are right, should be boolean :(

PR #502, opened 2019-05-06
Template text editor | placeholder

Yossitrx:
## Status 
**READY**

## Description
added Storybook support
Placeholder content editable as false. 

BOB-STYLE

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.stories.ts
+++ b/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.stories.ts
@@ -1,12 +1,5 @@
 import { storiesOf } from '@storybook/angular';
-import {
-  text,
-  select,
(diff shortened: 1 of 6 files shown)


projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.stories.ts
```diff
@@ -98,6 +92,25 @@ const note = `
...
+      options: [
+        {
+          value: 'First name',
+          id: '/root/firstName',
+          selected: false,
+        },
+        {
+          value: 'work | title',
+          id: '/work/title',
+          selected: false,
```
tomersela: @Yossitrx this shouldn't be part of the metadata this component gets.
It can be an internal state we generate inside the component, but not something our component API should expose.

PR #505, opened 2019-05-07
avatar size enum with numbers + css variable

galloween:


galloween:
@danazelniker this is the PR with the avatar size enum

projects/ui-framework/src/lib/services/utils/functional-utils.ts
```diff
@@ -2,6 +2,10 @@ export const keysFromArrayOrObject = (smth: string[] | {}): string[] => {
   return Array.isArray(smth) ? smth : Object.keys(smth);
 };
 
+export const getKeyByValue = (object: object, value: any) => {
+  return Object.keys(object).find(key => object[key] === value);
+};
+
```
omrihecht: no need for this

PR #508, opened 2019-05-08
added add file component

ronyrosenberg:
Added Add file component.


projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="container" [ngClass]="{'edit-mode': imageUrl}" (mouseover)="updateIconClass($event)">
```
omrihecht: i would imagine the icon class is a boolean, we can try a different approach and not go through the controller.
have
```
(mouseover)="iconHover=true"
(mouseleave)="iconHover=false"
```
in the b-icon:
```
[color]="iconHover ? iconColor.xxx : iconColor.yyy"
```

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.html
```diff
@@ -0,0 +1,8 @@
...
+  <b-icon [icon]="selectedIcon"
+          size="large"
```
omrihecht: use the enums.
if we change values this will break.
in the controller have
```
readonly iconSize = IconSize
```
then it's available as enum in the html

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.html
```diff
@@ -0,0 +1,8 @@
...
+          color="normal"
+          [color]="iconColor">
```
omrihecht: you have 2 colors attributes, remove the one that doesn't use an enum

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
```diff
@@ -0,0 +1,52 @@
...
+    align-items: center;
+    justify-content: center;
+    background-color: $grey-200;
+    position: relative;
+    &:hover {
+      background-color: $grey-300;
+    }
+
+    b-icon {
+      padding-left: 5px;
```
omrihecht: no need for this, if it centers the icon, it means the icon is no good and need to be replaced

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
```diff
@@ -0,0 +1,52 @@
...
+      background-color: $white;
+      border-style: solid;
+      &:hover {
+        background-color: rgba(83,83,83,0.5);
+        b-icon {
+          opacity: 1;
+          z-index: 999;
+          //.mat-icon svg g path {
+          //  fill: white !important;
+          //}
```
omrihecht: remove

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
```diff
@@ -0,0 +1,52 @@
...
+      background-color: $grey-300;
+    }
+
+    b-icon {
+      padding-left: 5px;
+      padding-bottom: 2px;
+    }
+    &.edit-mode {
+      b-icon {
+        opacity: 0;
```
omrihecht: you can add transition on the opacity with 0.2s

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
```diff
@@ -0,0 +1,52 @@
...
+      }
+      background-color: $white;
+      border-style: solid;
+      &:hover {
+        background-color: rgba(83,83,83,0.5);
```
omrihecht: perhaps would look nice to add transition to this property as well

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
```diff
@@ -0,0 +1,52 @@
...
+        }
+      }
+    }
+    .file-image {
+      background-size: cover;
+      background-repeat: no-repeat;
+      background-position: center;
+      position: absolute;
+      height: 50%;
+      width: 50%;
```
omrihecht: why 50%?

omrihecht: would probably be better to add the background image to the `container` element and remove the file-image.
will make the dom cleaner

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.ts
```diff
@@ -0,0 +1,30 @@
...
+export class AddFileComponent implements OnChanges {
+  @Input() icon: Icons;
+  @Input() imageUrl?: string;
+  @Output() clicked: EventEmitter<void> = new EventEmitter<void>();
+  selectedIcon: string;
+  iconColor: string;
+  constructor() {
+  }
+  ngOnChanges(): void {
+    this.selectedIcon = this.imageUrl ? 'pencil_icon' : this.icon;
```
omrihecht: you can move this logic to the html

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  @HostListener('click', ['$event'])
+  onClick($event) {
+    this.clicked.emit($event);
```
omrihecht: this event is the mouse event, do you need it?
better emit empty if there is no use for this.

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  }
+  updateIconClass() {
+    if (this.imageUrl) {
+      this.iconColor = 'white';
```
omrihecht: move logic to html

projects/ui-framework/src/lib/icons/icons.enum.ts
```diff
@@ -64,4 +64,6 @@ export enum Icons {
   success = 'success_icon',
   warning = 'warning_icon',
   no_folder = 'no_folder',
+  add_photo = 'add_photo',
+  pencil_icon = 'pencil_icon',
```
omrihecht: there is already a pencil icon, please make sure you are only adding new icons.

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.html
ronyrosenberg: 👍

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
galloween: also should be a variable - rgba($grey-700,0.5)

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
```diff
@@ -0,0 +1,57 @@
...
+        transition: background-color 0.1s ease-out;
+      }
+      &:hover:after {
+        opacity: 0.5;
+      }
+      background-size: cover;
+      background-repeat: no-repeat;
+      background-position: center;
+      background-origin: content-box;
+      padding: 15px;
```
omrihecht: this will not look good if the image has a background (not png with transparent background), you can tell ben about it

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+    it('Should add background image url to style if backgroundImage', () => {
+      component.imageUrl = 'url';
+      fixture.detectChanges();
+      const elementContainerStyle =
+        fixture.debugElement.query(By.css('.container')).nativeElement.getAttribute('style');
+      expect(elementContainerStyle).toEqual('background-image: url("url");');
+    });
+  });
+});
```
omrihecht: nice

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  styleUrls: ['./add-file.component.scss']
+})
+export class AddFileComponent {
+  @Input() icon: Icons;
+  @Input() imageUrl?: string;
+  @Output() clicked: EventEmitter<void> = new EventEmitter<void>();
+  readonly iconColor = IconColor;
+  readonly iconSize = IconSize;
+  readonly iconShape = Icons;
+  iconHover: boolean;
```
omrihecht: whats this for?

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.module.ts
```diff
@@ -0,0 +1,12 @@
...
+import { AddFileComponent } from './add-file.component';
+import {IconsModule} from '../../icons/icons.module';
+
+@NgModule({
+  declarations: [AddFileComponent],
+  imports: [CommonModule, IconsModule],
+  exports: [AddFileComponent],
+})
+export class AddFileModule {
+}
```
omrihecht: you need to export the module and component in the public_api file

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.ts
ronyrosenberg: removing

projects/ui-framework/src/lib/buttons-indicators/add-file/add-file.component.scss
ronyrosenberg: @levben we need to make a call here. Not sure there is a fit for every use case.

PR #513, opened 2019-05-12
quill scss imported to style

omrihecht:
fix quill import bug
image

@galloween , apparently we cannot have subfolders in files we import to style.scss (bob-style) so i moved all the quill styles to a single file in the root level of style. i checked, it looks ok, but i saw there were some things that were commented out, if they are not needed, ignore this message, otherwise please add them to quill.scss


galloween:
how do I reproduce this error? (i tried compiling both bob-style and hibob-web-client with no problems)
It's ok to have a million levels of folders in SCSS usually. if there is a problem with it, its somewhere else. maybe there is a syntax error or it doesnt like './' before folder...

ishaiborovoy:
@galloween it's not working when it is in shared / on-boarding the subfolder not exist

galloween:
well then its a misconfiguration in the package command:

if you change "*.*" to "**/*" on line 16 in package.json - it should fix it
@omrihecht @ishaiborovoy 

Screen Shot 2019-05-12 at 10 58 07

ishaiborovoy:
@galloween then change it if it fix it...

galloween:
yes, added it, version 2114 compiles fine now, the folder is copied.

PR #581, opened 2019-05-27
Truncate-tooltip optimization

galloween:
some optimizations towards more css-based solution

### DIFF
--- a/projects/ui-framework/src/lib/services/truncate-tooltip/truncate-tooltip.component.ts
+++ b/projects/ui-framework/src/lib/services/truncate-tooltip/truncate-tooltip.component.ts
@@ -52,8 +52,8 @@ export class TruncateTooltipComponent
     this.setMaxLines(value);
   }
   @Input() delay = 300;
-  @Input() expectChanges = true;
-  @Input() useCssVars = false;
+  @Input() expectChanges = false;
+  @Input() trustCssVars = false;
 
   private resizeSubscription: Subscription;
   private textContainer: HTMLElement;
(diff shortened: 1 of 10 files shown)


projects/ui-framework/src/lib/services/truncate-tooltip/truncate-tooltip.component.ts
```diff
@@ -63,18 +52,25 @@ export class TruncateTooltipComponent
     this.setMaxLines(value);
   }
   @Input() delay = 300;
+  @Input() expectChanges = true;
+  @Input() useCssVars = false;
```
omrihecht: I think the default values should be the more optimised option, meaning the user should input if he expectsChanges or not, and if using the CssVars is better performance, maybe this should be the default. wdyt?

projects/ui-framework/src/lib/style/functions.scss
```diff
@@ -0,0 +1,40 @@
...
+        @if $get == null {
+          @return null;
+        }
+      } @else {
+        @return null;
+      }
+    }
+  }
+  @return $get;
+}
```
omrihecht: what does this return?

galloween: this function returns deep-nested key
used like  map-get-deep($props, 'text' 'text-slot-1' 'font')
= lodash  get($props, '.text.textSlot1.font', null)

its not used in tooltip, but in avarar.

projects/ui-framework/src/lib/services/truncate-tooltip/truncate-tooltip.component.ts
galloween: then it means we need to enable the typography mixins to insert variables everywhere.
if somebody is not using mixins or setting font size manually, like font-size: 12px, it wont work...

PR #603, opened 2019-05-29
Pavel/truncate tooltip optimize

galloween:
- lazy init of matTooltip - after hovering for 200ms
- pure CSS tooltip mode - good for short tooltips

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte-core/rte-utils.service.ts
+++ b/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte-core/rte-utils.service.ts
@@ -16,7 +16,7 @@ export class RteUtilsService {
   constructor(private DOM: DOMhelpers) {}
 
   getHtmlContent(editor: Quill): string {
-    return !editor.root.innerText.trim() ? '' : editor.root.innerHTML;
+    return !editor.root.textContent.trim() ? '' : editor.root.innerHTML;
   }
 
   getTextLength(editor: Quill): number {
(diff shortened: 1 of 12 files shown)


projects/ui-framework/src/lib/form-elements/rich-text-editor/rte-core/rte-utils.service.ts
```diff
@@ -16,7 +16,7 @@ export class RteUtilsService {
   constructor(private DOM: DOMhelpers) {}
 
   getHtmlContent(editor: Quill): string {
-    return !editor.root.innerText.trim() ? '' : editor.root.innerHTML;
+    return !editor.root.textContent.trim() ? '' : editor.root.innerHTML;
```
omrihecht: whats the difference between innerText and textContent?

galloween: innertext is better but expensive. it calculates css and causes a reflow. but will return text exactly as visible (if text-transform was set to uppercase, it will be in uppercase)
textcontent will also return text that is inside script and style tags, and maybe even comments... but its very performant.
i think its safe to use textcontent when we dont expect crazy stuff and also on things that happen often (like in example above - getHtmlContent is called every keystroke).
for other cases when performance is not an issue, innerText is better...

omrihecht: 👍

galloween: innertext is accurate, textcontent is fast

PR #604, opened 2019-05-30
missing exports + 'changed' event for Chips Input

galloween:
- some interfaces where not exported...
- Revi Friedman requested change event output for Chip Input component


galloween:
this component doesnt have tests... i didnt get around to that yet.

PR #605, opened 2019-05-30
alert component and service

tomeraronovsky:
alert component and service
animations & new close button next prs


projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
```diff
@@ -0,0 +1,39 @@
+@import '../../style/colors';
+@import '../../style/typography';
+
+::ng-deep {
```
ishaiborovoy: I think better to use host too, see other components.  So the case will be isolated

omrihecht: @tomeraronovsky , ::ng-deep in this way is exposed to all always, this shouldn't be anywhere that is not supposed to shared everywhere.
good catch @ishaiborovoy

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
```diff
@@ -0,0 +1,39 @@
...
+  .b-alert {
+    background-color: $white;
+    border-radius: 4px;
+    box-shadow: 0 5px 10px 0 rgba(83,83,83,0.2);
+    min-width: 395px;
+    min-height: 78px;
```
omrihecht: i don't think you need min-height.
height should be determined by content
also width should be set as a const (or enum),  min-width could also mean the content can be 100vw, i don't think it's what they meant.
please use mixin for box-shadow (look for it in the code), i think there is also one for border-radius

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.html
```diff
@@ -0,0 +1,10 @@
+<ng-template #alertTemplateRef>
+  <div class="alert-container">
+    <b-icon class="close-icon" [icon]="closeIcon" [color]="closeColor" [size]="closeSize"></b-icon>
```
omrihecht: use button where the element has button behaviour

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
```diff
@@ -0,0 +1,39 @@
...
+
+    .alert-container {
+      position: relative;
+      display: flex;
+      flex-direction: row;
+      align-items: center;
```
omrihecht: no need to center align vertically, add padding to top and bottom instead

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.ts
```diff
@@ -0,0 +1,65 @@
+import {Component, TemplateRef, ViewChild, ViewContainerRef} from '@angular/core';
+import {cloneDeep, Dictionary} from 'lodash';
+import {IconColor, Icons, IconSize} from '../../icons/icons.enum';
+import {InfoStripIcon} from '../info-strip/info-strip.types';
+import {AlertConfig} from './alert.interface';
+import {Overlay, OverlayConfig, OverlayRef} from '@angular/cdk/overlay';
+import {TemplatePortal} from '@angular/cdk/portal';
+
+const ALERT_DURATION = 3000;
```
omrihecht: i think it's 7000, check hibob-web popup service (`showPositiveMessage`)

projects/ui-framework/src/lib/buttons-indicators/alert/alert.enum.ts
```diff
@@ -0,0 +1,6 @@
+export enum AlertType {
+  information = 'information',
+  warning = 'warning',
+  error = 'error',
+  success = 'success'
+}
```
omrihecht: 👍

projects/ui-framework/src/lib/buttons-indicators/alert/alert.module.ts
```diff
@@ -0,0 +1,20 @@
...
+import {AlertService} from './alert-service/alert.service';
+import {BrowserAnimationsModule, NoopAnimationsModule} from '@angular/platform-browser/animations';
+
+@NgModule({
+  declarations: [AlertComponent],
+  entryComponents: [AlertComponent],
+  imports: [
+    CommonModule,
+    IconsModule,
+    NoopAnimationsModule,
```
omrihecht: why is this here?
`NoopAnimationsModule`

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
tomeraronovsky: I wrapped the b-alert with ng-deep because b-alert is the panelClass, and I saw that every panelClass we wrap with ng-deep

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.html
tomeraronovsky: as we talked I need to change some settings for the X button, i will add it next pr, for now i gonna remove it

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
tomeraronovsky: didn't understand why, if you align it vertically it ensure it will looks good in the container

projects/ui-framework/src/lib/buttons-indicators/alert/alert.module.ts
tomeraronovsky: wrong, remove it

projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
```diff
@@ -0,0 +1,56 @@
...
+  private overlayConfig: OverlayConfig;
+  private templatePortal: TemplatePortal;
+  public overlayRef: OverlayRef;
+
+  constructor(private componentFactoryResolver: ComponentFactoryResolver,
+              private injector: Injector,
+              private overlay: Overlay) { }
+
+  public showAlert(config: AlertConfig): void {
+    this.injectAlertComponent();
```
omrihecht: 👍

projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  private injectAlertComponent(): void {
+    const componentFactory = this.componentFactoryResolver.resolveComponentFactory(AlertComponent);
+    this.alertComponentRef = componentFactory.create(this.injector);
+  }
+
+  private getConfig(): OverlayConfig {
+    const positionStrategy = this.overlay.position().global()
+      .centerHorizontally().top('20px');
+    const panelClass = 'b-alert';
```
omrihecht: rename `b-alert-panel`, the class will be added to the material generated panel, not the b-alert component

projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    const backdropClass = '';
+    return {
+      disposeOnNavigation: true,
+      hasBackdrop: false,
+      backdropClass,
```
omrihecht: if you don't have backdrop, no need for backdrop class

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.html
```diff
@@ -0,0 +1,9 @@
+<ng-template #alertTemplateRef>
+  <div class="alert-container">
+    <b-icon class="icon" [icon]="iconsDic[alertConfig.alertType].icon" [color]="iconsDic[alertConfig.alertType].color" [size]="iconSize"></b-icon>
```
omrihecht: why do you need ng-template?
also you can remove alert-container, and use the host as the container

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
```diff
@@ -0,0 +1,42 @@
+@import '../../style/colors';
+@import '../../style/typography';
+@import '../../style/variables';
+
+$alert_width: 395px;
+
+::ng-deep {
+  .b-alert {
```
omrihecht: b-alert-panel

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
omrihecht: please remove height,
use shadow mixin,
remove min-width (should be width, f i remember correctly, you can add to the panel config)
if you want let's have a short session and do the css right

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.scss
omrihecht: because the container should be determined by the content height, not the other way around

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Component({
+  selector: 'b-alert',
+  templateUrl: './alert.component.html',
+  styleUrls: ['./alert.component.scss'],
+})
+export class AlertComponent {
+  @ViewChild('alertTemplateRef') alertTemplateRef: TemplateRef<any>;
+  public alertConfig: AlertConfig;
+  public closeAlertCallback: Function;
```
omrihecht: doesn't look like should be public

tomeraronovsky: it must be public for binding it in the service

projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      const ALERT_DURATION_TICK = 7001;
+      const alertConfig: AlertConfig = {
+        alertType: AlertType.success,
+        text: 'text',
+        title: 'title'
+      };
+      alertService.showAlert(alertConfig);
+      tick(ALERT_DURATION_TICK);
+      expect(alertService.overlayRef.dispose).toHaveBeenCalled();
+    }));
```
omrihecht: great!

omrihecht: it's good practice to have tests for every public method, you can add a small one to `closeAlert`

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      alertType: AlertType.success,
+      text: 'text',
+      title: 'title'
+    };
+    fixture.detectChanges();
+  });
+
+  it('should call to close alert callback', () => {
+    component.closeAlertCallback = jasmine.createSpy();
+    component.closeAlert();
```
omrihecht: a better test would be to click the close button, not invoke the method

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      providers: [
+        { provide: AlertService, useValue: {} },
+      ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(AlertComponent);
+    component = fixture.componentInstance;
```
omrihecht: add `spyOn(component, 'closeAlertCallback');`

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    component.alertConfig = {
+      alertType: AlertType.success,
+      text: 'text',
+      title: 'title'
+    };
+    fixture.detectChanges();
+  });
+
+  it('should call to close alert callback', () => {
+    component.closeAlertCallback = jasmine.createSpy();
```
omrihecht: remove

projects/ui-framework/src/lib/buttons-indicators/alert/alert.component.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    component.closeAlert();
+    expect(component.closeAlertCallback).toHaveBeenCalled();
+  });
+});
```
omrihecht: you should also test the config to dom implementation (the integration test)

projects/ui-framework/src/public_api.ts
```diff
@@ -81,6 +81,10 @@ export { LinkColor, LinkTarget } from './lib/buttons-indicators/link/link.enum';
 // InfoTooltip Module
 export { InfoTooltipModule } from './lib/popups/info-tooltip/info-tooltip.module';
 export { InfoTooltipComponent } from './lib/popups/info-tooltip/info-tooltip.component';
+// Alert Module
+export { AlertService } from './lib/buttons-indicators/alert/alert-service/alert.service';
+export { AlertModule } from './lib/buttons-indicators/alert/alert.module';
+export { AlertConfig } from './lib/buttons-indicators/alert/alert.interface';
```
omrihecht: you also need to export the AlertComponent, even though it's part of the module

omrihecht:
you need to update/rebase to merge

PR #609, opened 2019-06-02
enhancements

omrihecht:
- added padding to tabs
 - updated close button icon
 - remove hover/active from tertiary types (making it "ghost")
 - update dialog style
 - updated typography colors
 - added keys to tabs and menu

closes hibobio/bob-style/issues/596
closes hibobio/bob-style/issues/592
closes hibobio/bob-style/issues/589
closes hibobio/bob-style/issues/598

### DIFF
--- a/projects/ui-framework/src/lib/navigation/tabs/tabs.component.html
+++ b/projects/ui-framework/src/lib/navigation/tabs/tabs.component.html
@@ -4,8 +4,10 @@
                [selectedIndex]="selectedIndex"
                [disableRipple]="true">
 
-  <mat-tab *ngFor="let tab of tabs"
-           label="{{tab.label}}">
+  <mat-tab *ngFor="let tab of tabs">
+    <ng-template mat-tab-label>
+      <span class="{{tab?.key}}">{{tab.label}}</span>
+    </ng-template>
   </mat-tab>
 
 </mat-tab-group>
--- a/projects/ui-framework/src/lib/navigation/tabs/tabs.component.scss
+++ b/projects/ui-framework/src/lib/navigation/tabs/tabs.component.scss
@@ -38,7 +38,7 @@
   .mat-tab-label {
     height: 30px;
     min-width: 0;
-    padding: 0 10px;
+    padding: 0 15px;
     color: $grey-700;
     opacity: 1;
     text-transform: uppercase;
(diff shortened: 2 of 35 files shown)


projects/ui-framework/src/lib/navigation/tabs/tabs.component.html
```diff
@@ -4,8 +4,10 @@
                [selectedIndex]="selectedIndex"
                [disableRipple]="true">
 
-  <mat-tab *ngFor="let tab of tabs"
-           label="{{tab.label}}">
+  <mat-tab *ngFor="let tab of tabs">
```
galloween: i found out there is a better way to use tabs. material has a separate component that provides same behaviour as tabs header - mat-tab-nav-bar. Last example here: https://material.angular.io/components/tabs/examples
so we can get just the navigation, without the (empty) panels!

something like this:
```
<nav mat-tab-nav-bar
     #tabGroup
     [ngClass]="'tabs-'+type"
     (selectedTabChange)="onSelectChange($event)"
     disableRipple="true">
  <a mat-tab-link
     *ngFor="let tab of tabs"
     (click)="activeLink = tab"
     [active]="activeLink == tab"
     [ngClass]="tab?.key"
     disableRipple="true">
    {{tab.label}}
  </a>
</nav>
```

projects/ui-framework/src/lib/navigation/tabs/tabs.component.scss
```diff
@@ -38,7 +38,7 @@
   .mat-tab-label {
     height: 30px;
     min-width: 0;
-    padding: 0 10px;
+    padding: 0 15px;
```
galloween: this will result in padding of 25px on left and right
is that what you wanted?
(10 pixels is added via js - I set inline width css of width + 10 - to get rid of a jump when font weight changes)

projects/ui-framework/src/lib/navigation/tabs/tabs.component.html
galloween: nope. mat-tab-nav-bar doesnt have right-left scroll buttons... :(
so staying with mat-tab-group

projects/ui-framework/src/lib/navigation/tabs/tabs.component.scss
omrihecht: i am increasing the space between tabs from 20px to 30px

galloween: it was 30 before. so now its 40

omrihecht: ok, it was to close before according to raanan, i played with the padding when i sat with him

omrihecht: `(10 pixels is added via js - I set inline width css of width + 10 - to get rid of a jump when font weight changes)`
nice!!

PR #625, opened 2019-06-04
add animation fade in fade out to alert

tomeraronovsky:
add animation fade in fade out to alert


tomeraronovsky:
I close this one, will do the animation in different pr

PR #630, opened 2019-06-05
add icons

benbaler:
can you add screenshot, i think this could also just be a simple chevron like we have in css

https://hibob.invisionapp.com/d/main#/console/14664741/342603215/preview


benbaler:
@omrihecht yes it is cheveron, but i need it as icon.

PR #633, opened 2019-06-06
dialog title fix

benbaler:
closes https://github.com/hibobio/hibob/issues/24954

### DIFF
--- a/projects/ui-framework/src/lib/popups/dialog/dialog.component.scss
+++ b/projects/ui-framework/src/lib/popups/dialog/dialog.component.scss
@@ -33,6 +33,9 @@ $footerSize: 50px;
         div[b-dialog-sub-title] {
           margin-top: 5px;
         }
+        b-display-2 {
+          overflow-wrap: break-word;
+        }
       }
       .mat-dialog-content {
         padding: 0 $dialogPadding 20px;


projects/ui-framework/src/lib/popups/dialog/dialog.component.scss
```diff
@@ -33,6 +33,11 @@ $footerSize: 50px;
         div[b-dialog-sub-title] {
           margin-top: 5px;
         }
+        b-display-2 {
```
omrihecht: not sure about this
who requested it?

benbaler: here: https://github.com/hibobio/hibob/issues/24954

omrihecht:
@benbaler , i see,
the solution should be to break the word, not limit all titles to one line with ellipsis.

galloween:
> @benbaler , i see,
> the solution should be to break the word, not limit all titles to one line with ellipsis.

@benbaler @omrihecht 

add
overflow-wrap: break-word;
to break long words. i used the same solution in tooltip for similar 'bug'

benbaler:
@omrihecht @galloween done

projects/ui-framework/src/lib/popups/dialog/dialog.component.scss
```diff
@@ -33,6 +33,11 @@ $footerSize: 50px;
...
+          overflow: hidden;
+          text-overflow: ellipsis;
+          overflow-wrap: break-word;
```
omrihecht: remove:
```
overflow: hidden;
text-overflow: ellipsis;
```

PR #637, opened 2019-06-11
side menu three dots fix

benbaler:
when one item selected and you hover on the next one you will see '...' three dots, when you click on that three dots the panel opened and the three dots disappear, it shouldn't. i fixed it so it will not disappear.

### DIFF
--- a/projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.html
+++ b/projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.html
@@ -12,9 +12,8 @@
          *ngIf="option.actions?.length"
          [ngStyle]="{'opacity': (showActions || selected) ? 1 : 0}">
         <b-menu [menu]="option.actions"
-                (click)="stopBubbling($event)"
-                (mouseleave)="stopBubbling($event); setShowActions(true)"
-                (closeMenu)="showActions = false">
+                (click)="stopBubbling($event); setShowActions(true, 25)"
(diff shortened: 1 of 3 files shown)


projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.html
```diff
@@ -12,9 +12,8 @@
          *ngIf="option.actions?.length"
          [ngStyle]="{'opacity': (showActions || selected) ? 1 : 0}">
         <b-menu [menu]="option.actions"
-                (click)="stopBubbling($event)"
-                (mouseleave)="stopBubbling($event); setShowActions(true)"
-                (closeMenu)="showActions = false">
+                (click)="stopBubbling($event); setShowActions(true, 25)"
```
danazelniker: what is this number? 25? just random timeout?

omrihecht: i'm not sure i understand this pr

benbaler: yes

benbaler: i want it to happen after the next call so it will stay true... no other elegant way to do that...

omrihecht:
@benbaler , please add description, i don't understand what this pr is for, is this a bug fix? what's the bug?

PR #640, opened 2019-06-12
employees showcase animation

benbaler:
http://css3.bradshawenterprises.com/cfimg/


projects/ui-framework/src/lib/buttons-indicators/avatar/avatar.component.scss
```diff
@@ -22,7 +22,7 @@
   position: relative;
   border-radius: 50%;
   background-color: $grey-200;
-  transition: box-shadow 0.2s, transform 0.2s;
```
benbaler: not sure about replacing that, what it actually doing???

omrihecht: this is transition of the avatar when hovering, why are you removing this?

PR #641, opened 2019-06-12
fix alert bug

tomeraronovsky:
* open only one alert each time
* clear timer reference on close

### DIFF
--- a/projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
+++ b/projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
@@ -12,23 +12,28 @@ export class AlertService {
   private overlayConfig: OverlayConfig;
   private templatePortal: TemplatePortal;
   public overlayRef: OverlayRef;
+  public isOpen: boolean;
+  private timeRef: NodeJS.Timer;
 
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
```diff
@@ -12,23 +12,28 @@ export class AlertService {
   private overlayConfig: OverlayConfig;
   private templatePortal: TemplatePortal;
   public overlayRef: OverlayRef;
+  public isOpen: boolean;
+  private timeRef: NodeJS.Timer;
```
omrihecht: really????

projects/ui-framework/src/lib/buttons-indicators/alert/alert-service/alert.service.ts
```diff
@@ -50,5 +55,7 @@ export class AlertService {
 
   public closeAlert(): void {
     this.overlayRef.dispose();
+    this.isOpen = false;
+    clearTimeout(this.timeRef);
   }
```
omrihecht: cool

tomeraronovsky:
@omrihecht yes, the return type of setTimeout is NodeJS.Timer, but if I used window.setTimeout the return type was number

PR #657, opened 2019-06-17
Pavel/lightbox

galloween:
Lightbox that can
- show image
- show url in iframe
- show component (via component-renderer)

Note:
- no tests


projects/ui-framework/src/lib/popups/lightbox/lightbox.component.html
```diff
@@ -0,0 +1,30 @@
...
+         border="0"
+         alt="">
+  </div>
+
+  <iframe class="lightbox-iframe"
+          *ngIf="config.url && !config.image"
+          [src]="sanitizer.bypassSecurityTrustResourceUrl(config.url)"
+          frameborder="0"
+          allowfullscreen>
+  </iframe>
```
doroncy: i am afraid of security issues due to this dynamic iframe src option. not sure its worth it, we can stay with image and component options.

projects/ui-framework/src/lib/popups/lightbox/lightbox.component.scss
```diff
@@ -0,0 +1,148 @@
...
+  }
+}
+@keyframes fade-in {
+  0% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
+}
```
doroncy: i think we should have some global animation key frames  so we would not need to rewrite and invent it all over again, looks like fade-in, fade-out, zoom-out can be generic

projects/ui-framework/src/lib/popups/lightbox/lightbox.service.spec.ts
```diff
@@ -0,0 +1,126 @@
...
+//         title: 'title'
+//       };
+//       lightboxService.showLightbox(lightboxConfig);
+//       tick(ALERT_DURATION_TICK);
+//       expect(lightboxService.overlayRef.hostElement).toBeNull();
+//       expect(lightboxService.overlayRef.hasAttached()).toBeFalsy();
+//       expect(lightboxService.isOpen).toBeFalsy();
+//     }));
+//   });
+// });
```
doroncy: need to update tests

doroncy:
@galloween there are many files and also many changes here, like avatar, alerts, lightbox and other stuff. please try to do more PRs with less changes in each of them, even a 1 file PR is great. also each PR sholud be attached to an issue which this pr closes.

projects/ui-framework/src/lib/popups/lightbox/lightbox.component.html
galloween: yes. there is now our own url-sanitizer service that (currently) only allows youtube & vimeo links (later can be expanded).
the bypassSecurityTrustResourceUrl is removed from the lightbox.

projects/ui-framework/src/lib/popups/lightbox/lightbox.component.scss
galloween: and we do have already. but what is somebody changes them or forgets to import...
this way the component is self-contained.
but it can be a mixin. so that the source code is in one place.

projects/ui-framework/src/lib/popups/alert/alert-service/alert.service.ts
```diff
@@ -54,7 +54,9 @@ export class AlertService {
   }
 
   public closeAlert(): void {
-    this.overlayRef.dispose();
+    if (this.overlayRef) {
+      this.overlayRef.dispose();
+    }
```
omrihecht: this should have test

projects/ui-framework/src/lib/popups/lightbox/lightbox.service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+        this.lightboxPortal = new ComponentPortal(LightboxComponent);
+        this.lightboxComponentRef = this.overlayRef.attach(this.lightboxPortal);
+        this.lightboxComponentRef.instance.closeLightboxCallback = bind(
+          this.closeLightbox,
+          this
+        );
+      }
+
+      this.lightboxComponentRef.instance.ngOnChanges({
```
omrihecht: should be returned

projects/ui-framework/src/lib/popups/lightbox/lightbox.service.ts
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      config = {
+        image: config.image && this.url.validateImg(config.image),
+        video: config.video && this.url.domainAllowed(config.video as string),
+        component: config.component && config.component,
+        fillScreen: config.fillScreen
+      };
+
+      if (!this.isOpen) {
+        this.overlayRef = this.overlay.create(this.overlayConfig);
```
omrihecht: let's create a new reference for every time we open lightbox, and return it to the consumer

projects/ui-framework/src/lib/popups/lightbox/lightbox.interface.ts
```diff
@@ -0,0 +1,19 @@
...
+  image?: string;
+  video?: string | SafeResourceUrl;
+  fillScreen?: boolean;
+}
+
+export interface LightboxData {
+  overlayRef: OverlayRef;
+  lightboxComponentRef: ComponentRef<LightboxComponent>;
+  config?: LightboxConfig;
+}
```
omrihecht: only the RenderedComponent is exported to the user right?

galloween: also the lightboxComponentRef - for no real reason

galloween: and the processed config (with safe urls)

projects/ui-framework/src/lib/popups/lightbox/lightbox.service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  private overlayConfig: OverlayConfig = {
+    disposeOnNavigation: true,
+    hasBackdrop: false,
+    panelClass: 'b-lightbox-panel',
+    positionStrategy: this.overlay.position().global(),
+    scrollStrategy: this.overlay.scrollStrategies.block()
+  };
+
+  public showLightbox(config: LightboxConfig): LightboxData {
```
omrihecht: i think this should return the overlay ref (for example can be closed by the consumer, or listen to close)

omrihecht: never mind, i saw it includes it...

projects/ui-framework/src/lib/popups/lightbox/lightbox.interface.ts
galloween: the RenderedComponent is not exported actually

PR #659, opened 2019-06-18
table secondary style

omrihecht:
- added secondary style to data table
 - change grid column gap to 25px

fyi: @benbaler

### DIFF
--- a/projects/ui-framework/src/lib/style/grid-layout.scss
+++ b/projects/ui-framework/src/lib/style/grid-layout.scss
@@ -5,7 +5,7 @@ $breakpoint-device-lg: 1200px;
 @mixin gridLayout12Cols() {
   display: grid;
   grid-template-columns: repeat(12, 1fr);
-  grid-column-gap: 12px;
+  grid-column-gap: 24px;
 }
 
 .grid-layout-12-cols {
--- a/projects/ui-framework/src/lib/table/table/table.component.spec.ts
+++ b/projects/ui-framework/src/lib/table/table/table.component.spec.ts
@@ -1,11 +1,4 @@
-import {
-  async,
-  ComponentFixture,
-  fakeAsync,
-  flush,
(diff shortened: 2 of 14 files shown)


projects/ui-framework/src/lib/style/grid-layout.scss
```diff
@@ -5,7 +5,7 @@ $breakpoint-device-lg: 1200px;
 @mixin gridLayout12Cols() {
   display: grid;
   grid-template-columns: repeat(12, 1fr);
-  grid-column-gap: 12px;
+  grid-column-gap: 25px;
```
benbaler: 24px?

benbaler: @raananmorag

projects/ui-framework/src/lib/table/table/table.component.spec.ts
```diff
@@ -88,6 +80,12 @@ describe('TableComponent', () => {
       const agRoot = shadowRoot.querySelector('.ag-root') as HTMLElement;
       expect(getComputedStyle(agRoot).maxHeight).toEqual('450px');
     });
+    it('should add class "secondary" if type is secondary', () => {
+      component.type = TableType.Secondary;
+      fixture.autoDetectChanges();
```
benbaler: i think you can use autoDetectChanges once at the start of the testing file and it will detect changes automatically, right?
in this case detectChanges is enough i guess...

omrihecht: if you want to check component inputs, you want to keep the detectChanges after you update the inputs

projects/ui-framework/src/lib/style/grid-layout.scss
omrihecht: good catch! it is 24

Issue #662, opened 2019-06-18
Multi Select


omrihecht:
- Add list number in multi select when truncating
![image](https://user-images.githubusercontent.com/24462119/59700186-85da3600-91fb-11e9-80ac-c70495c15306.png)
 - Add prefix (for employee ref)

PR #664, opened 2019-06-18
Pavel/form elements update

galloween:
- rte lists fix https://github.com/hibobio/hibob/issues/25430
- form elements margins (no margins if only child)
- multi-select values counter + input prefix  #662

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
+++ b/projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
@@ -18,14 +18,6 @@ $input-focus-border-color: $grey-700;
   &:only-child {
     margin: 0;
   }
-
-  &:first-child {
-    margin-top: 0;
-  }
-
-  &:last-child {
-    margin-bottom: 0;
-  }
 }
 
 @mixin transfer-margins($from: 'b-input') {
--- a/projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
+++ b/projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
@@ -12,8 +12,25 @@ $label-hint-margin: 6px;
 $input-border-color: $grey-500;
 $input-focus-border-color: $grey-700;
 
-@mixin transfer-margins($from: 'b-input') {
+@mixin form-element-margins {
   margin: $form-element-margin;
+
+  &:only-child {
+    margin: 0;
+  }
+
+  &:first-child {
+    margin-top: 0;
+  }
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+}
+
+@mixin transfer-margins($from: 'b-input') {
+  @include form-element-margins;
+
   &::ng-deep {
     #{$from} {
       margin: 0;
@@ -77,9 +94,9 @@ $input-focus-border-color: $grey-700;
   display: flex;
   width: 100%;
   flex-direction: column;
-  margin: $form-element-margin;
   box-sizing: border-box;
   text-align: left;
+  @include form-element-margins;
 
   // wrap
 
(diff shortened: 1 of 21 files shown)


projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
```diff
@@ -12,8 +12,25 @@ $label-hint-margin: 6px;
...
 
-@mixin transfer-margins($from: 'b-input') {
+@mixin form-element-margins {
   margin: $form-element-margin;
+
+  &:only-child {
+    margin: 0;
+  }
+
+  &:first-child {
```
omrihecht: this doesn't look right to me, what if there are 2 elements in one row? first child has margin, second doesn't?

projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
```diff
@@ -12,8 +12,25 @@ $label-hint-margin: 6px;
...
+    margin-top: 0;
+  }
+
+  &:last-child {
+    margin-bottom: 0;
+  }
```
omrihecht: same, i think only-child is the only one we need

projects/ui-framework/src/lib/form-elements/form-element.mixin.scss
galloween: you are very right

projects/ui-framework/src/lib/form-elements/lists/multi-select/multi-select.component.html
```diff
@@ -18,6 +25,10 @@
   </span>
 
   <div class="bfe-suffix">
+    <span *ngIf="truncate.tooltipEnabled"
+          class="total-values">
+      ({{selectedValuesMap.length}})
+    </span>
```
omrihecht: can this be tested?

projects/ui-framework/src/lib/form-elements/lists/multi-select/multi-select.component.ts
```diff
@@ -94,6 +103,13 @@ export class MultiSelectComponent extends BaseSelectPanelElement
     }
   }
 
+  ngAfterViewInit(): void {
+    setTimeout(() => {
+      this.showPrefix =
+        this.hasPrefix || this.prefix.nativeElement.childNodes.length > 0;
```
omrihecht: i remember i had a problem with empty tags like angular comments, there was no problem?

projects/ui-framework/src/lib/form-elements/lists/multi-select/multi-select.component.html
galloween: i tested it! empirically

projects/ui-framework/src/lib/form-elements/lists/multi-select/multi-select.component.ts
galloween: yes there may be actually, event if there is a space.
i have a better test in DOMutils service. i used that before, but in refactor replaced with this simplified one because it was a headache to pass the service to parent abstracts etc. maybe I should bring it back, or convert the domUtils to a collection of constants maybe?

projects/ui-framework/src/lib/form-elements/lists/multi-select/multi-select.component.html
omrihecht: empirically? this is too big of a word for me ;)
but it would be nice to add a test that the number in brackets are displayed

galloween: there is a test for tooltip in single/multi-select components (shows on long text, doesnt by default).
so if tooltip is enabled, its enabled property can be trusted to be true, i guess...

galloween: we can add test for the number, for sure

galloween: empirically = by observation, rather than science ;)

PR #665, opened 2019-06-19
side menu three dots fix

benbaler:
https://github.com/hibobio/hibob/issues/25586

just fix for simple open close menu, without the solution for pressing sharing and etc.

### DIFF
--- a/projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.ts
+++ b/projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.ts
@@ -33,6 +33,7 @@ export class SideMenuOptionComponent implements OnChanges {
   public icons = Icons;
   public colors = IconColor;
   public types = ButtonType;
+  public menuOpened: boolean;
 
   public showActions = false;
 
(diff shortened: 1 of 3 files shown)


projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.ts
```diff
@@ -75,8 +76,8 @@ export class SideMenuOptionComponent implements OnChanges {
     $event.stopPropagation();
   }
 
-  setShowActions(status: boolean, timeout: number = 0): void {
-    setTimeout(() => this.showActions = status, timeout);
+  setShowActions(status: boolean): void {
```
danazelniker: do you still use this function? I see you remove the call from the html

benbaler: yes i use it in the html in other places

projects/ui-framework/src/lib/navigation/side-menu/side-menu-option/side-menu-option.component.ts
```diff
@@ -75,8 +76,8 @@ export class SideMenuOptionComponent implements OnChanges {
...
+    setTimeout(() => this.showActions = status);
```
danazelniker: check if you need timeout

benbaler: you right

PR #670, opened 2019-06-19
better RTE controls management

galloween:
- if no controls/disableControls input is provided, use defaults
- controls can be enabled/disabled dynamically

And while I was there:
- Added label above RTE. the label inside is now placeholder. if only 1 of 2 is present, it will be treated as placeholder (for backwards compatibility).
- fixed panels width (kind of)
- moved chars counter to toolbar suffix


galloween:
> we can add a test for the placeholder + label and merge

added tests, will merge when green

PR #675, opened 2019-06-20
empty state component

tomeraronovsky:
empty state component

https://hibob.invisionapp.com/d/main#/console/14664741/369585405/preview

### DIFF
--- a/projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.html
+++ b/projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.html
@@ -1,0 +1,13 @@
+<div class="b-empty-state">
+  <b-icon class="empty-state-icon"
+          [icon]="icon"
+          [color]="iconColor"
+          [size]="iconSize"></b-icon>
+
+  <p class="empty-state-text">{{text}}</p>
+
+  <b-button *ngIf="buttonLabel"
+            (clicked)="onClick()"
+            [type]="buttonType"
+            [size]="buttonSize">{{buttonLabel}}</b-button>
+</div>
--- a/projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.html
+++ b/projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.html
@@ -1,13 +1,11 @@
-<div class="b-empty-state">
-  <b-icon class="empty-state-icon"
-          [icon]="icon"
-          [color]="iconColor"
-          [size]="iconSize"></b-icon>
+<b-icon class="empty-state-icon"
+        [icon]="config.icon"
+        [color]="iconColor"
+        [size]="iconSize"></b-icon>
 
-  <p class="empty-state-text">{{text}}</p>
+<p class="empty-state-text">{{config.text}}</p>
 
-  <b-button *ngIf="buttonLabel"
-            (clicked)="onClick()"
-            [type]="buttonType"
-            [size]="buttonSize">{{buttonLabel}}</b-button>
-</div>
+<b-button *ngIf="config.buttonLabel"
+          (clicked)="buttonClick.emit()"
+          [type]="buttonType"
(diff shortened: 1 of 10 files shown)


projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="b-empty-state">
```
omrihecht: can you use the host instead of wrapper div?

projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.ts
```diff
@@ -0,0 +1,25 @@
...
+
+@Component({
+  selector: 'b-empty-state',
+  templateUrl: './empty-state.component.html',
+  styleUrls: ['./empty-state.component.scss']
+})
+export class EmptyStateComponent {
+  @Input() public icon: Icons;
+  @Input() public text: string;
+  @Input() public buttonLabel: string;
```
omrihecht: i would move these into 1 interface, EmptyStateConfig or something like that

projects/ui-framework/src/lib/icons/icon.component.scss
```diff
@@ -164,5 +164,9 @@
       height: 46px;
       width: 46px;
     }
+    &.xx-large {
+      height: 60px;
+      width: 60px;
+    }
```
omrihecht: was this discussed with the designers?
are they aware that this component require another icon type?

projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.ts
```diff
@@ -0,0 +1,25 @@
...
+  @Output() buttonClick: EventEmitter<void> = new EventEmitter<void>();
+  readonly iconSize: IconSize = IconSize.xxLarge;
+  readonly iconColor: IconColor = IconColor.light;
+  readonly buttonType: ButtonType = ButtonType.secondary;
+  readonly buttonSize: ButtonSize = ButtonSize.medium;
+
+  constructor() { }
+
+  public onClick(): void {
+    this.buttonClick.emit();
```
omrihecht: you can do this from the html

projects/ui-framework/src/lib/icons/icon.component.scss
tomeraronovsky: this is the size they gave to the icon and we don't support it for now. i will let them know

projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.component.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../style/colors.scss';
+
+:host {
+  justify-content: center;
+  display: flex;
+  flex-direction: column;
+  .empty-state-text {
+    margin: 10px 0px 25px 0px;
```
omrihecht: 0 should not have px,
correct way to write this is `10px 0 25px`

projects/ui-framework/src/lib/buttons-indicators/empty-state/empty-state.stories.ts
```diff
@@ -0,0 +1,56 @@
...
+const EmptyStateStories = storiesOf(
+  ComponentGroupType.ButtonsAndIndicators,
+  module
+).addDecorator(withKnobs);
+
+const template =
+  `<b-empty-state [config]="config"
+                  (buttonClick)="buttonClicked($event)"></b-empty-state>`;
+
+const storyTemplate = `<b-story-book-layout [title]="'Empty State'">${template}</b-story-book-layout>`;
```
omrihecht: story template has `text-align: center`, did you check it looks the same outside story template?

tomeraronovsky: done

PR #691, opened 2019-06-24
Pavel/icons font

galloween:
**new icon system**
- total icons now 92 (vs ~50 before)
- uses icon font (vs svg elements)
- icon can also be applied just by css-class (b-icon component not necessary)

**possible breaking changes:**
- Icons enum changed values
- IconService removed
- material icon removed => elements/classes changed (test may break)

(tests in bob-style are fixed)

**need to update in hibob-web-client:**
- employee-details.component (uses strings not enums for icon)
- media-uploader.component (uses strings not enums for icon)


projects/ui-framework/src/lib/style/style.scss
```diff
@@ -9,6 +9,9 @@
 @import './tooltip';
 @import './line-clamp';
 @import './dialog';
+@import './icons';
+
+@import url('https://i.icomoon.io/public/fe7353d4db/UntitledProject/style.css');
```
omrihecht: move to icons.scss

PR #693, opened 2019-06-25
Pavel/cards update

galloween:
employee-card simple implementation, no tests


projects/ui-framework/src/lib/buttons-indicators/avatar/avatar.interface.ts
```diff
@@ -1,4 +1,17 @@
...
+
+export interface Avatar {
+  imageSource: string;
+  title?: string;
+  subtitle?: string;
+  caption?: string;
+  chip?: Chip;
+  badge?: AvatarBadge | BadgeConfig;
+  orientation?: AvatarOrientation;
+}
```
omrihecht: this is breaking changes? or are you leaving the old Input as they are?

galloween: no, no breaking changes. just an generic interface describing Avatar-component inputs, that can be used for component-renderer, or - in this case - the employee-card that has avatar inside it, so i wanted to mirror its interface

PR #700, opened 2019-06-27
single list hide none option

benbaler:
https://github.com/hibobio/hibob/issues/25654


benbaler:
@omrihecht sorry it is single select, but, why you think it is not a good solution? what are you suggesting? how should i remove the "- none -" option?

omrihecht:
I think you should not remove None option.
you should validate the form and request a value instead ("This field is required"), None for single select is the same as clear in multi-select, should always be there

doroncy:
I agree with @omrihecht  there is no real value in removing it.
anyways, just for future features, the APIs of components/functions should be clear and with the same pattern, so if this API have ```showSingleGroupHeader``` do not add ```hideNoneOption``` it complex the API, when sometimes true is for show and sometimes true, is for hiding.

PR #712, opened 2019-07-01
employees showcase click on avatar

benbaler:
https://github.com/hibobio/hibob/issues/25959


projects/ui-framework/src/lib/buttons-indicators/employees-showcase/employees-showcase.component.html
```diff
@@ -3,7 +3,9 @@
     <b-avatar class="employees-showcase-avatar"
               *ngIf="i < avatarsToFit - 1 || (i < avatarsToFit && !showMore) || (i < avatarsToFit && last && showMore)"
               [imageSource]="employee.imageSource"
-              [size]="avatarSize"></b-avatar>
+              [size]="avatarSize"
+              [isClickable]="true"
```
danazelniker: This should be true only if the user passes a `clicked` function

benbaler: i use this solution to check that
https://stackoverflow.com/questions/33441393/is-there-a-way-to-check-for-output-wire-up-from-within-a-component-in-angular-2

projects/ui-framework/src/lib/buttons-indicators/employees-showcase/employees-showcase.component.ts
```diff
@@ -158,7 +159,14 @@ export class EmployeesShowcaseComponent
 
   private switchEmployeesImage(firstIndex, secondIndex) {
     const firstEmployeeImage = this.employees[firstIndex].imageSource;
+    const firstEmployeeId = this.employees[firstIndex].id;
```
danazelniker: use _.assign

PR #727, opened 2019-07-07
Pavel/chip-list, mocks

galloween:
- better chip list keyboard control
- methods for easier mock data


projects/ui-framework/src/lib/chips/multi-list-and-chips/multi-list-and-chips.component.ts
```diff
@@ -68,7 +68,7 @@ export class MultiListAndChipsComponent implements OnChanges {
             key: group.key,
             name: group.groupName
           },
-          type: ChipType.info
+          type: this.chipListConfig.type !== ChipType.avatar && ChipType.info
```
omrihecht: remove all group selected check in case of avatar

PR #731, opened 2019-07-07
Pavel/utils service upd

galloween:
hostlistener('document.click') and hostlistener('document.keydown') were triggering page change detection. replaced with a service rxjs observable (does not trigger chnage detection)

### DIFF
--- a/projects/ui-framework/src/lib/popups/lightbox/lightbox.component.ts
+++ b/projects/ui-framework/src/lib/popups/lightbox/lightbox.component.ts
@@ -12,6 +12,7 @@ import { ButtonType } from '../../buttons-indicators/buttons/buttons.enum';
 import { LightboxConfig } from './lightbox.interface';
 import { UtilsService } from '../../services/utils/utils.service';
 import { Subscription } from 'rxjs';
+import { filter } from 'rxjs/operators';
 import { Keys } from '../../enums';
 import { isKey } from '../../services/utils/functional-utils';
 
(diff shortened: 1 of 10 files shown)


projects/ui-framework/src/lib/popups/panel/panel.component.ts
```diff
@@ -50,21 +53,30 @@ export class PanelComponent implements OnDestroy {
...
     this.mouseLeaveDebounce = debounce(this.closePanel, HOVER_DELAY_DURATION);
   }
 
+  ngOnInit(): void {
+    this.windowKeydownSubscriber = this.utilsService
+      .getWindowKeydownEvent()
+      .subscribe(event => {
+        if (isKey(event.key, Keys.escape)) {
+          this.closePanel();
+        }
```
omrihecht: should be .pipe(filter)

projects/ui-framework/src/lib/popups/lightbox/lightbox.component.ts
```diff
@@ -39,8 +45,18 @@ export class LightboxComponent implements OnChanges {
     );
   }
 
-  @HostListener('document:keydown.escape') handleEscape() {
-    this.closeLightboxCallback();
+  ngOnInit(): void {
+    this.windowKeydownSubscriber = this.utilsService
+      .getWindowKeydownEvent()
+      .subscribe(event => {
+        if (isKey(event.key, Keys.escape)) {
```
omrihecht: use rx filter

PR #775, opened 2019-07-16
Pavel/outside-zone eventManager plugin

galloween:
- outside-zone eventManager plugin
- stories update that adds the stats component globally

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.module.ts
+++ b/projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.module.ts
@@ -13,8 +13,7 @@ import { DOMhelpers } from '../../services/utils/dom-helpers.service';
 import { PlaceholderRteConverterService } from './rte-placeholder/placeholder-rte-converter.service';
 import { SingleListModule } from '../lists/single-list/single-list.module';
 import { InputMessageModule } from '../input-message/input-message.module';
-import { EVENT_MANAGER_PLUGINS } from '@angular/platform-browser';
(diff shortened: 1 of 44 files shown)


projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.component.html
```diff
@@ -7,7 +7,7 @@
   <div #quillEditor
        [attr.data-length]="hasSuffix && (minChars > 0 || maxChars) ? ( (length || 0) + (maxChars ? '/' + maxChars : '')) :
        null"
-       (keydown)="$event.stopPropagation()"
+       (keydown.outside-zone)="$event.stopPropagation()"
```
omrihecht: we need to check this doesn't break the surveys, maybe we need to update the surveys when we upgrade bob-style

projects/ui-framework/src/lib/form-elements/rich-text-editor/rte.module.ts
```diff
@@ -28,6 +29,11 @@ import { InputMessageModule } from '../input-message/input-message.module';
     InputMessageModule
   ],
   exports: [RichTextEditorComponent],
-  providers: [RteUtilsService, DOMhelpers, PlaceholderRteConverterService]
+  providers: [
+    RteUtilsService,
+    DOMhelpers,
+    PlaceholderRteConverterService,
+    EventManagerPlugins[0]
```
omrihecht: array?

PR #786, opened 2019-07-18
chip list radio buttons + option for specific selected index.

harellevy:


projects/ui-framework/src/lib/chips/chip-list/chip-list.component.ts
```diff
@@ -32,6 +32,8 @@ export class ChipListComponent implements OnChanges {
 
   @ViewChildren('list') public list: QueryList<ChipComponent>;
 
+  @Input() radioSelect = false;
+  @Input() activeIndex: number = null;
```
galloween: should be selectedIndex maybe? (if we use 'selected' elsewhere)

projects/ui-framework/src/lib/chips/chip-list/chip-list.component.ts
```diff
@@ -80,13 +82,23 @@ export class ChipListComponent implements OnChanges {
     }
   }
 
+  checkSelected(chip, index) {
+    if (this.radioSelect) {
+      return this.activeIndex === index;
+    } else {
+      return chip.selected;
+    }
```
omrihecht: return ```
this.radioSelect
? this.activeIndex===index
: chip.selected;
```

File: projects/ui-framework/src/lib/chips/chip-list/chip-list.component.ts [hunk 3]
```diff
@@ -80,13 +82,23 @@ export class ChipListComponent implements OnChanges {
...
+
   onChipRemove(chip: Chip): void {
     this.removed.emit(chip);
   }
 
-  onChipClick(event: MouseEvent, chip: Chip): void {
+  onChipClick(event: MouseEvent, chip: Chip, index: number): void {
     event.stopPropagation();
-    if (this.config.selectable) {
+    if (this.radioSelect) {
```
galloween: lets move this logic to selectChip method because its called from multiple places (including keyboard navigation)

omrihecht:
moved to https://github.com/hibobio/bob-style/pull/790

PR #802, opened 2019-07-21
add group key to getSelected function

danazelniker:


projects/ui-framework/src/lib/form-elements/lists/list-change/list-change.ts
```diff
@@ -22,18 +22,22 @@ export class ListChange {
...
+            map(group.options, option => ({ option,  groupName: group.groupName, key: group.key }))
           ),
         [],
       )
       .filter(groupOption => groupOption.option.selected)
-      .map(groupOption => ({ id: groupOption.option.id, groupName: groupOption.groupName }))
+      .map(groupOption => ({
+        id: groupOption.option.id,
+        groupName: groupOption.groupName,
+        groupKey: groupOption.key
```
omrihecht: this could be undefined, groupOption.key is optional if i remember

projects/ui-framework/src/lib/form-elements/lists/list.interface.ts
```diff
@@ -40,3 +40,9 @@ export interface ListFooterActions {
   cancel?: boolean;
   apply?: boolean;
 }
+
+export interface GroupOption {
+  id: number | string;
+  groupName: string;
+  groupKey?: string | number;
+}
```
omrihecht: We are still in early stages, maybe we should consider updating the interface to something scalable, i think atm only @danidin uses it.
@danidin , wdyt?

projects/ui-framework/src/lib/form-elements/lists/list-change/list-change.ts
danazelniker: if its undefined then groupKey will be undefined also. should I set it to null instead?

omrihecht: i think undefined is ok also, it indicates it doesn't exist, we need to think of someone consuming it, what is clearer, i'm not sure.
on seconds thoughts maybe null is more consistent with our patterns

projects/ui-framework/src/lib/form-elements/lists/list.interface.ts
danazelniker: @omrihecht maybe we can keep the same structure as `SelectGroupOption` and include only the selected option under the `options` key. @danidin WDYT?

danidin: I'm open to any suggestions, can make the appropriate changes in places where I'm using it.
@omrihecht what would you change in the interface?

danazelniker: @danidin @omrihecht So I'm going with keeping `SelectGroupOption` interface. I'll create another function and mark Dani's as deprecated.

omrihecht: cool, let's make sure to remove the deprecated method as soon as possible

PR #847, opened 2019-07-31
Table - fix max height

rinat-shemer:
use ```max-height```

### DIFF
--- a/projects/ui-framework/src/lib/table/table/styles/table.component.scss
+++ b/projects/ui-framework/src/lib/table/table/styles/table.component.scss
@@ -22,9 +22,12 @@
   }
 }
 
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/table/table/styles/table.component.scss
```diff
@@ -29,7 +29,7 @@
     display: block;
     width: 100%;
     height: 100%;
-    max-height: 450px;
+    max-height: var(--max-height);
```
galloween: I suggest you add on the host the defaut variable value

:host ::ng-deep {
  display: block;
--max-height: 450px;

then it will be overwritten by the JS inline style

PR #848, opened 2019-07-31
high charts wrapper and example component pie chart. no tests yet.

harellevy:


projects/ui-framework/src/lib/charts/chart/chart-core.ts
```diff
@@ -39,7 +38,7 @@ export abstract class ChartCore implements AfterViewInit, OnInit, OnChanges {
     '#FF9655',
     '#FFF263',
     '#6AF9C4'];
-  @Input() height = 700;
+  @Input() height = 500;
```
omrihecht: we should default to the same size as the home page pie charts, 150x150

projects/ui-framework/src/lib/charts/donut-chart-text/donut-chart-text.stories.ts
```diff
@@ -16,9 +16,9 @@ const template = `
     [name]="name"
     [donutInnerSize]="donutInnerSize"
   >
-    <b-display-4>
+    <b-display-3 style="color: #9d9c9c;">
       {{text}}
-    </b-display-4>
+    </b-display-3>
```
omrihecht: nice!

projects/ui-framework/src/lib/charts/pie-chart/pie-chart.stories.ts
```diff
@@ -47,16 +47,17 @@ const note = `
...
+  donut | boolean | make pie chart donut chart | false
+  showDataLabels | boolean | shows label in pie | false
+  legend | boolean | shows legend | false
+  colorPalette | string[] | color palette array | default array of colors
+  height | number | height of chart | 500
+  name | string | name of series | none
+  donutInnerSize | number | defining the inner white circle in a donut pie chart | 60
+  donutWidth | number | overrides donutInnerSize by applying width of donut instead inner circle width | none
+  title | string | title of chart | none
+  pointFormat | string | tooltip formatter | {series.name}: <b>{point.percentage:.1f}%</b>
```
omrihecht: i think a lot of the inputs can be merged into an interface,
so should be something like:
```
dataConfig: {
name: string
data: (type?)
pointFormat: string (default to ...)
}
pieConfig: {
donut: boolean (default=true)
showLabels: boolean (default=false)
showLegend: boolean (default=false)
height: number (default=150)
donutWidth: number (default 35)
title: string (default to null)
}
```

donutInnerSize can be removed if we are overriding it with donutWidth, we should keep one or the other

projects/ui-framework/src/lib/charts/chart/chart-core.ts
harellevy: I changed default value for pie chart and not all charts.

projects/ui-framework/src/lib/charts/pie-chart/pie-chart.component.ts
```diff
@@ -0,0 +1,75 @@
...
+  @Input() name: string;
+  @Input() donut = false;
+  @Input() donutInnerSize = 60;
+  @Input() donutWidth: number;
+  constructor(
+    public zone: NgZone,
+    public cdr: ChangeDetectorRef
+  ) {
+    super(zone, cdr);
+    this.height = 150;
```
harellevy: this is what i did instead. initialise pie chart with its own default height on contractor.

projects/ui-framework/src/lib/charts/pie-chart/pie-chart.component.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as Highcharts from 'highcharts';
+
+export class MockNgZone extends NgZone {
+  run(fn: Function): any {
+    return fn();
+  }
+}
+
+
+fdescribe('PieChartComponent', () => {
```
omrihecht: remove

PR #861, opened 2019-08-05
Harel/donut chart center, transparent background

harellevy:


harellevy:
@omrihecht , @doroncy I merge it, nothing complicated, just small css issues.

PR #888, opened 2019-08-14
- emoji initial commit

harellevy:
- emoji mapping by category, name.
- emoji picker with custom button, click outside close (outside of zone)
- some tests


projects/ui-framework/src/lib/form-elements/emoji/emoji.component.html
```diff
@@ -0,0 +1,25 @@
+<div class="emoji-panel" *ngIf="emojiMenuState" (click)="stopPropagationClick($event)">
+  <div class="emoji-header">
+    <b-display-3 class="emoji-title">Add reaction</b-display-3>
```
doroncy: bob style does not have translation, therefor all the texts should be input values for the components

projects/ui-framework/src/lib/form-elements/emoji/emoji.component.html
```diff
@@ -0,0 +1,25 @@
...
+      </div>
+    </div>
+  </div>
+  <div class="emoji-footer">
+    <div class="icon" *ngFor="let emoji of commonEmojis" (click)="selectEmoji(emoji)">
+      {{emoji}}
+    </div>
+  </div>
+</div>
+<div class="button-wrapper" (click)="stopPropagationClick($event); toggleMenu()">
```
doroncy: what is this wrapper? and why stop prop. ?

harellevy: the wrapper is for the button to activate the emoji panel, the stop propagation is when you click outside of panel it should close, but when click on the button / panel of emoji it shouldn't close.

projects/ui-framework/src/lib/form-elements/emoji/emoji.component.html
harellevy: ok i'll add these as inputs

harellevy:
also need to do more mapping for the emojis for current way of reading them and mapping server data that is saved to my mapping. i'm working on it

projects/ui-framework/src/lib/form-elements/emoji/emoji.component.scss
```diff
@@ -0,0 +1,127 @@
+@import "../../style/colors";
+$emoji-padding: 15px;
+:host {
+  position: relative;
+  display: inline-block;
+}
+.emoji-panel {
```
galloween: we have special mixins for panels - so that they look the same.
import popups/overlay.scss

then you can check panel.component.scss for example of use.

basically, you put
@include OverlayPanel;
@include OverlayPointerHolders;
@include OverlayPositionBounding('center');
and also
@include OverlayPositionBounding('above'); - for the arrow

projects/ui-framework/src/lib/form-elements/emoji/emoji.component.html
galloween: you can also put (click.outside-zone) here, because you have detectchanges in toggleMenu

PR #898, opened 2019-08-15
chain select: refactor delete button

danidin:


danidin:
@galloween This is much better! Thanks!

PR #899, opened 2019-08-19
Danny/chain select structural directive

danidin:
Modify Chain Select component to use structural directive instead of Component Renderer.

Main Change:

Chain Select is a component that accepts any Component and creates a list to which the user can add or remove copies of the input component.

The first version of Chain Select uses Component Renderer to handle the input component.
This PR does it with a structural directive.

The pros of this approach:
- No need for input configuration for the component, everything is done in the mark up.
- No need to use component renderer

Secondary Change:

Chain Select used to hold a state of all select elements inside of it.
This was a leaky abstraction, since Chain Select couldn't know how to translate its input state into select output.
The new version holds no state and emits single change events plus index, instead of a list of changes.

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
@@ -2,8 +2,7 @@
 
   <div class="chain-link-row">
 
-    <b-component-renderer [render]="chainLink.selectComponentConfig">
-    </b-component-renderer>
+    <ng-container *ngTemplateOutlet="contentChild.tpl"></ng-container>
 
     <b-square-button class="delete-button"
                      *ngIf="chainLinkList.length > 1"
--- a/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
@@ -6,7 +6,7 @@
     </ng-container>
 
     <b-square-button class="delete-button"
-                     *ngIf="chainLinkList.length > 1"
+                     *ngIf="selectedItemList.length > 1"
                      [type]="buttonType.tertiary"
                      [size]="buttonSize.large"
                      [icon]="icons.delete"
--- a/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
@@ -2,7 +2,7 @@
 
   <div class="chain-link-row">
 
-    <ng-container *ngTemplateOutlet="contentChild.tpl; context: { selected: selected, selectChange: handleChange.bind(this), index: i }">
+    <ng-container *ngTemplateOutlet="contentChild?.tpl; context: { selected: selected, selectChange: handleChange.bind(this), index: i }">
     </ng-container>
 
     <b-square-button class="delete-button"
--- a/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
@@ -1,8 +1,9 @@
-<ng-container *ngFor="let chainLink of chainLinkList; index as i;">
+<ng-container *ngFor="let selected of selectedItemList; index as i;">
 
   <div class="chain-link-row">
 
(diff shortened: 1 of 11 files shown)


projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
```diff
@@ -1,12 +1,12 @@
-<ng-container *ngFor="let chainLink of chainLinkList; index as i;">
+<ng-container *ngFor="let selected of selectedItemList; index as i;">
 
   <div class="chain-link-row">
 
-    <b-component-renderer [render]="chainLink.selectComponentConfig">
-    </b-component-renderer>
+    <ng-container *ngTemplateOutlet="contentChild?.tpl; context: { selected: selected, selectChange: handleChange.bind(this), index: i }">
```
itayod:
```suggestion
    <ng-container *ngTemplateOutlet="contentChild?.tpl;
       context: { selected: selected, selectChange: handleChange.bind(this), index: i }"
     >
```

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
```diff
@@ -1,12 +1,12 @@
...
+    </ng-container>
 
     <b-square-button class="delete-button"
-                     *ngIf="chainLinkList.length > 1"
+                     *ngIf="selectedItemList.length > 1"
```
itayod:
```suggestion
                     *ngIf="selectedItemList.?length > 1"
```

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.ts
```diff
@@ -1,78 +1,41 @@
...
-        }
-      }));
-      this.state = this.selectComponentConfig.selectedIds.map(() => null);
-    }
-  }
+  @ContentChild(ChainSelectDirective, { static: true }) contentChild !: ChainSelectDirective;
 
   public addChainLink() {
-    this.chainLinkList.push(this.createEmptyChainLink(this.chainLinkList.length));
+    this.selectedItemList.push(null);
```
itayod: IMHO it should be output since there will be many use cases for implementing this event... (e.g default value or auto select the new item, autofocus....)

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.ts
```diff
@@ -1,78 +1,41 @@
...
   styleUrls: ['./chain-select.component.scss']
 })
-export class ChainSelectComponent implements OnInit {
+export class ChainSelectComponent {
   @Input() actionLabel: string;
-  @Input() selectComponent: Type<any>;
-  @Input() selectComponentConfig: SelectComponentConfig;
-  @Output() selectChange: EventEmitter<ListChange[]> =
-    new EventEmitter<ListChange[]>();
+  @Input() selectedItemList: any[] = [null];
```
itayod: why null?

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.ts
```diff
@@ -1,78 +1,41 @@
...
-        component: this.selectComponent,
-        attributes: {
-          ...this.selectComponentConfig.filterFn && { filterFn: this.selectComponentConfig.filterFn },
-        },
-        handlers: {
-          [this.selectComponentConfig.outputKey]: $event => this.handleChange($event, index)
-        }
-      }
-    };
+    this.selectedItemList.splice(index, 1);
```
itayod: normally you want the state to be immutable (better for change detections) I would simply use a filter and send the list  value

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.ts
danidin: yeah, I think I'll add it

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.ts
danidin: start with one empty select by default

projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.ts
danidin: oh snap! Fixing...

PR #909, opened 2019-08-21
support micro avatar size for notifications (30px)

harellevy:
micro avatar size for notification message avatar.


harellevy:
This is the size from the new notification panel design (30pxl. Its a different design from the old one. Maybe we need to take it offline to the design team.

omrihecht:
not used

PR #921, opened 2019-08-28
fix bug with position which is not fixed. panel wasn't centered

harellevy:


harellevy:
fix bug with position which is not fixed. panel wasn't centered in case of position static, relative, absolute.

PR #928, opened 2019-09-01
max-width (for long emoji picker title)

harellevy:


omrihecht:
@harellevy , if you need this merge when green

PR #934, opened 2019-09-02
update collapsible style

omrihecht:


omrihecht:
Basically i need the title to have border bottom, i am making it sticky in the employee profile, thats the reason for the change

galloween:
@omrihecht 
so it wont ever be collapsible, right?
then, the least-changes solution can be (box shadow will overlay the panel border-top, but will look like border when pos fixed):

`
.bcp-section:not(.bcp-section-collapsible),
.bcp-section.bcp-section-expanded {
  .bcp-header {
    box-shadow: 0 1px 0 0 $grey-500;	
  }
}
`

if it should also be collapsible, then 
`
.bcp-section:not(.bcp-section-collapsible),
.bcp-section.bcp-section-expanded {
  .bcp-header {
    box-shadow: 0 1px 0 0 $grey-500;
  }
}
`

plus background-color white on .bcp-header  in general, i would imagine

galloween:
also, if you would like to make the sticky headers an option of the collapsible section, there is a  [options] input that takes an object where you can add that option.....

galloween:
added the box-shadow to https://github.com/hibobio/bob-style/pull/935
(which also fixes the overflow issue in date range picker)

omrihecht:
@galloween fixed in different pr

PR #949, opened 2019-09-08
emoji-chip-tooltip

harellevy:
- comment input autofocus


projects/ui-framework/src/lib/chips/emoji-chip-list/emoji-chip-list.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="emoji-chip-item" *ngFor="let chip of chips"
-     (click)="chipClick(chip)">
+     (click)="chipClick(chip)" [matTooltip]="chip.tooltip" matTooltipPosition="above">
```
omrihecht: this is a bit expensive, maybe try to use [tooltipData], it's lighter/faster

PR #953, opened 2019-09-09
- optional name, date

harellevy:


harellevy:
no need. the interface is fine, Omri fixed it on web-client

PR #954, opened 2019-09-09
Adding divided flag and title color to the collapsible section compon…

avgdorom:
…ent.

### DIFF
--- a/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section-example.module.ts
+++ b/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section-example.module.ts
@@ -38,6 +38,7 @@ import { InputModule } from '../../form-elements/input/input.module';
       [collapsible]="collapsible"
       [expanded]="expanded"
       [disabled]="disabled"
+      [divided]="divided"
       (opened)="onOpened()"
       (closed)="onClosed()"
     >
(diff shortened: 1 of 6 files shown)


projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section-example.module.ts
```diff
@@ -337,6 +340,7 @@ export class CollapsibleSectionExample2Component implements OnChanges {
   @Input() disabled = false;
 
   @Input() title = mockText(randomNumber(2, 5));
+  @Input() titleColor = '#5555ff';
```
doroncy: I dont want to have default color like this, please remove this and try to make the ngStyle works when no value in titleColor

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section-example.module.ts
```diff
@@ -337,6 +340,7 @@ export class CollapsibleSectionExample2Component implements OnChanges {
   @Input() disabled = false;
 
   @Input() title = mockText(randomNumber(2, 5));
```
doroncy: also remove the mocktext it looks like pasha forgot it here

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section-example.module.ts
avgdorom: But it's just for the example in Storybook. There's no default in the actual component.

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section-example.module.ts
avgdorom: What should be here?

PR #962, opened 2019-09-15
Menu changes

rinat-shemer:
- check if there is action before calling it
- sent the child on action click


projects/ui-framework/src/lib/navigation/menu/menu.component.ts
```diff
@@ -37,8 +37,10 @@ export class MenuComponent implements OnChanges {
   }
 
   onClick($event, child): void {
-    child.action($event);
-    this.actionClick.emit();
+    if (child.action) {
+      child.action($event);
+    }
+    this.actionClick.emit(child);
```
omrihecht: please explain

PR #963, opened 2019-09-15
comment with linkify, export linkify pipe, highlight pipe

harellevy:


projects/ui-framework/src/lib/comments/comment-list/comment-item/comment-item.component.scss
```diff
@@ -25,6 +25,9 @@
     }
     .comment-body {
       line-height: 20px;
+      ::ng-deep a {
+        color: $primary-500 !important;
```
omrihecht: remove important

PR #964, opened 2019-09-15
Pavel/daterange

galloween:
- updated dte-fns to v2

### DIFF
--- a/projects/ui-framework/package.json
+++ b/projects/ui-framework/package.json
@@ -15,7 +15,7 @@
     "ag-grid-community": "^20.0.0",
     "ag-grid-enterprise": "^20.0.0",
     "highcharts": "^7.1.2",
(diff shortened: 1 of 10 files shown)


projects/ui-framework/package.json
```diff
@@ -15,7 +15,7 @@
     "ag-grid-community": "^20.0.0",
     "ag-grid-enterprise": "^20.0.0",
     "highcharts": "^7.1.2",
-    "date-fns": "^1.30.1"
+    "date-fns": "^2.2.1"
```
omrihecht: this might have breaking changes, let's leave it as is, and update in separate pr

galloween: this is a separate PR for update ;)
date-fns is only used in datepicker, so should not break anything

PR #965, opened 2019-09-15
chain select: add static mode

danidin:
Add ability to lock the chain to the initial input state, only allowing the user to change the pre-existing select element values.


projects/ui-framework/src/lib/form-elements/lists/chain-select/chain-select.component.html
```diff
@@ -15,13 +15,14 @@
 
   </div>
 
-  <div class="then">
+  <div class="then" [ngClass]="{ 'then--static': staticMode }">
```
galloween: you can also add  'last as last' (or let last = last) to *ngFor and then use *ngIf="staticMode && last" here. may be easier to see whats going on this way.

PR #1011, opened 2019-10-02
- emoji fix multiple add emoji (after close wasn't opening panel)

harellevy:


harellevy:
closes https://github.com/hibobio/hibob/issues/28634

PR #1029, opened 2019-10-16
add content footer to single list and single select

danazelniker:


projects/ui-framework/src/lib/form-elements/lists/single-list/single-list.stories.ts
```diff
@@ -26,6 +26,9 @@ const template = `
 <b-single-list [options]="options"
                (selectChange)="selectChange($event)"
                [showSingleGroupHeader]="showSingleGroupHeader">
+               <div single-list-footer style="display: flex; justify-content: flex-end; padding: 15px 15px;">
```
galloween: this should be in .footer styles in single-list. only text button (with footer attribute) should be passed.

also I suggest:
```
align-items: center;
padding: 0 15px;  
height: 50px;
```
instead of ```padding: 15px 15px``` - same as in list-footer we use for multi-list

projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts
```diff
@@ -36,6 +36,9 @@ const template = `
                  [errorMessage]="errorMessage"
                  [hintMessage]="hintMessage"
                  [showSingleGroupHeader]="showSingleGroupHeader">
+                 <div single-select-footer style="display: flex; justify-content: flex-end; padding: 15px 15px;">
```
galloween: also remove styles. and maybe lets use simpler [footer] for selector?

projects/ui-framework/src/lib/form-elements/lists/single-list/single-list.stories.ts
galloween: you also may need to add:
```
  border-bottom-left-radius: $border-radius;
  border-bottom-right-radius: $border-radius;
```

danazelniker: I don't think it should be in .footer styles. since it's ng-content anyone that uses this footer can use any style he wants.
why do you think I need the `border-bottom-left-radius` ? it looks good without it (I also tried with different background color

projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts
danazelniker: because we have 2 levels of ng-content I need to use different name for each one

projects/ui-framework/src/lib/form-elements/lists/single-list/single-list.stories.ts
galloween: precisely to not let people use any styles, but our design-system styles.
we have the list footer for multi-select, this should look the same (i would think).

<b-single-list> has overflow hidden, thats why you dont see the border angles sticking out. Its up to up you - I would add border-radius, to not depend on overflow.

galloween: and you definitely need to add background-color white by the way :)

Screen Shot 2019-10-16 at 10 07 45

projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts
galloween: not really, you can use the same - try it :)
```<ng-content footer select="[footer]"></ng-content>```

Comment (danazelniker) [reply projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts]: nice!

Comment (galloween) [reply projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts]: maybe you should actually just use the existing list-footer.component... it has an option to just have the 'clear' button (withou cancel-apply). you can modify it to also accept content, if somebody ever needs something other then clear, cancel, apply....

**Comment by danazelniker** [comment]
@galloween I'm done. did all the changes. still needed to add style in the story.

File: projects/ui-framework/src/lib/form-elements/lists/single-list/single-list.stories.ts [hunk 3]
```diff
@@ -26,6 +26,9 @@ const template = `
 <b-single-list [options]="options"
                (selectChange)="selectChange($event)"
                [showSingleGroupHeader]="showSingleGroupHeader">
+               <div footer style="display: flex; justify-content: flex-end; height: 100%">
```
galloween: ```display: flex; justify-content: flex-end;``` should go to the .footer, too :)

projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts
danazelniker: Actually I need `upload doc` btn to I don't think it should be in list-footer.component

projects/ui-framework/src/lib/form-elements/lists/single-list/single-list.stories.ts
danazelniker: I only need it in the end in my case... it could be anything

galloween: i see, then just the display: flex...

projects/ui-framework/src/lib/form-elements/lists/single-select/single-select.stories.ts
galloween: ... if only list-footer could take content! ;)

projects/ui-framework/src/lib/form-elements/lists/single-list/single-list.stories.ts
danazelniker: but I need it on the div also since it's the container of the btn....

galloween: you can put ```style="margin-left: auto"``` on the button to have the same effect (aligned to right).
sorry I have a point about unnecessary elements. you can leave the div, but there is no actual need.

danazelniker: Cool! It'll do that! thanks :)

danazelniker:
@galloween OK now I'm done :)

danazelniker:
@galloween OK. now I'm done :)

PR #1034, opened 2019-10-22
emoji picker - title as description, not as shortname

harellevy:


harellevy:
closes https://github.com/hibobio/bob-style/pull/1034

PR #1051, opened 2019-10-27
Adding an option to disable the OK button in a dialog

avgdorom:


avgdorom:
@omrihecht Ultimately, I'm working on another solution, but I still think it's a good thing to have the "disabled" option here. WDYT?

avgdorom:
Approved verbally 😄

PR #1075, opened 2019-11-05
add pre line to comments

omrihecht:
https://github.com/hibobio/hibob/issues/29624


omrihecht:
not deploying for some reason, moves here
https://github.com/hibobio/bob-style/pull/1076

PR #1076, opened 2019-11-05
add pre line to comments

omrihecht:
Support native line breaks in comments
https://github.com/hibobio/hibob/issues/29624


galloween:
what if user adds a million new lines? i suggest adding a pipe that 
```
.replace(/\n{2,}/g,'\n\n')
```

PR #1078, opened 2019-11-06
fix donut chart missing tooltips

harellevy:


harellevy:
closed https://github.com/hibobio/hibob/issues/29922

PR #1091, opened 2019-11-11
Pavel/lists upd

galloween:
- returned OnCancel multiSelect tests
- added some list footer actions tests
- hide group header checkbox when showing search results


omrihecht:
thanks

PR #1093, opened 2019-11-12
add toggle strategy to breadcrumbs

itayod:
Added toggle strategy to breadcrumbs which will allow displaying all titles (we need that for compensation management)


itayod:
one neat, we will need to create another indicator for the active step now... but there is no style for that yet

PR #1119, opened 2019-11-18
create basic list

danidin:
Basic List component to satisfy the need for basic lists with or without menu for each item in bob.

First usage can be seen here:
https://hibob.invisionapp.com/d/main#/console/17981764/390271881/preview

Both in E-sign and Read approval sections

### DIFF
--- a/projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
@@ -1,0 +1,12 @@
+<ol>
+  <li *ngFor="let item of items">
+    <b-icon [icon]="item.icon"></b-icon>
+    <b-body>{{item.label}}</b-body>
+    <b-menu *ngIf="menu" [menu]="menu">
+      <b-square-button menu-trigger
+                       type="tertiary"
+                       icon="b-icon-more-horiz">
+      </b-square-button>
+    </b-menu>
+  </li>
+</ol>
--- a/projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
@@ -1,8 +1,8 @@
 <ol class="bbl-list">
-  <li class="bbl-list__item" *ngFor="let item of items; index as i">
-    <b-icon class="bbl-list__item__icon" [icon]="item.icon"></b-icon>
-    <span class="bbl-list__item__label" data-max-lines="2">{{item.label}}</span>
-    <div class="bbl-list__item__action-container">
+  <li class="bbl-list-item" *ngFor="let item of items; index as i">
+    <b-icon class="bbl-list-icon" [icon]="item.icon"></b-icon>
+    <span class="bbl-list-label" data-max-lines="2">{{item.label}}</span>
+    <div class="bbl-list-action-container">
       <ng-container
         *ngTemplateOutlet="contentChild?.tpl; context: { item: item, index: i }">
       </ng-container>
--- a/projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
+++ b/projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
@@ -1,12 +1,11 @@
 <ol>
-  <li *ngFor="let item of items">
+  <li *ngFor="let item of items; index as i">
     <b-icon [icon]="item.icon"></b-icon>
     <b-body>{{item.label}}</b-body>
-    <b-menu *ngIf="menu" [menu]="menu">
-      <b-square-button menu-trigger
-                       type="tertiary"
-                       icon="b-icon-more-horiz">
-      </b-square-button>
-    </b-menu>
+    <div class="action-container">
+      <ng-container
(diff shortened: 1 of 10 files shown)


projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
```diff
@@ -0,0 +1,12 @@
+<ol>
+  <li *ngFor="let item of items">
+    <b-icon [icon]="item.icon"></b-icon>
+    <b-body>{{item.label}}</b-body>
```
galloween: b-body here doesnt give anything here, as its page default font size/weight/style already. a span will be more appropriate

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
```diff
@@ -0,0 +1,12 @@
+<ol>
```
galloween: kudos for using ol!

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  padding: 0;
+  margin: 0;
+  background-color: white;
+  border: 1px solid $grey-300;
+  border-radius: $border-radius;
+
+  li {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-start;
```
galloween: these rules dont change anything, as its the default flex behaviour, can be removed

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.scss
```diff
@@ -0,0 +1,31 @@
...
+    &:last-of-type {
+      border-bottom: none;
+    }
+
+    b-icon {
+      margin-right: 5px;
+    }
+
+    b-menu {
+      margin-left: auto;
```
galloween: bravo

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.stories.ts
```diff
@@ -0,0 +1,68 @@
...
+  ## Basic list
+
+  #### Module
+  *BasicListModule*
+
+  #### Properties
+  Name | Type | Description | Default value
+  --- | --- | ---
+  items | BasicListItem[] | List of items to display | none
+  menu | MenuItem[] | Optional menu for each list item | undefined
```
galloween: we now use this notation: [input], (output), so [items], [menu]; if no default value, put &nbsp; instead of none/undefined

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.module.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@NgModule({
+  declarations: [
+    BasicListComponent,
+  ],
+  imports: [
+    CommonModule,
+    IconsModule,
+    MenuModule,
+    TypographyModule,
```
galloween: no need for typography here

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
galloween: also, for all typography components, you can use class instead, for example <span class="b-body"> (only in bob-style, where we try to make the components as light as possible; in the app, using typography components is preferred)

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.ts
```diff
@@ -0,0 +1,18 @@
...
+  templateUrl: './basic-list.component.html',
+  styleUrls: ['./basic-list.component.scss']
+})
+export class BasicListComponent implements OnInit {
+  @Input() items: BasicListItem[];
+  @Input() menu?: MenuItem[];
+
+  constructor() { }
+
+  ngOnInit() {
```
galloween: let's remove empty method

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.scss
```diff
@@ -0,0 +1,31 @@
+@import '../../../style/common-imports';
+
+ol {
```
galloween: it's better to use classes:  .bbl-list for ol, .bbl-list-item for il (bbl as prefix that stands for b-basic-list).
also better practice to write css flat (put li, b-icon, b-menu declaration outside ol), it will make selectors less specific an will be easier to override from outside, if needed

projects/ui-framework/src/lib/form-elements/lists/basic-list/basic-list.component.html
galloween: also while you are at it, test what happens if you 1) put a single very long line (no white space), 2) very long text
it might overflow in an ugly way.
for safequarding, you can add `min-width: 0` on the b-body/span css and `data-max-lines="2"` on the html element (that will limit its height to 2 lines of text and truncate the rest)

danidin:
@galloween @omrihecht 
Fixed PR comments, Changed component API to receive any content for actions instead of just menu.

PR #1178, opened 2019-12-08
Pavel/ lists-selects selectedIDs order bug fixed

galloween:
should fix https://github.com/hibobio/hibob/issues/30709


ishaiborovoy:
Cool I'll test it on web-client

PR #1182, opened 2019-12-09
download option for charts

harellevy:


projects/ui-framework/bob-charts/src/charts/chart/chart-core.ts
```diff
@@ -51,6 +53,11 @@ export abstract class ChartCore implements AfterViewInit {
   @Input() height = 500;
   @Input() title: string = null;
   @Input() legend = false;
+  @Input() set exportChartType (type: ExportingMimeTypeValue) {
+    if (typeof type !== 'undefined') {
+      this.exportChart(type);
+    }
+  }
```
galloween: this exports the chart immeditately when input changes???

omrihecht: Yes, you pass 'SVG' and it starts download process

galloween: this should be described in the notes precisely (because its not the behaviour people would expect)

omrihecht: this would only work if you change, you can't download the same file type twice

projects/ui-framework/bob-charts/src/charts/chart/chart-core.ts
```diff
@@ -71,6 +72,13 @@ export abstract class ChartCore implements AfterViewInit {
     </div>`;
   }
 
+  exportChart(type: ExportingMimeTypeValue) {
+    (this.highChartRef as any).exportChart(
+      {
+        type: type
+      }, {chart: {backgroundColor: '#ffffff'}}
```
omrihecht: remove background

PR #1205, opened 2019-12-18
support tree able table

itayod:


package.json
```diff
@@ -47,7 +47,8 @@
     "quill": "^1.3.6",
     "rxjs": "~6.5.2",
     "tributejs": "^4.0.0",
-    "zone.js": "~0.9.1"
+    "zone.js": "~0.9.1",
+    "@ag-grid-enterprise/all-modules": "latest"
```
ishaiborovoy: Not sure latest is the best thing here... it can break things if they update there version when we are not ready yet...

itayod: you are right, I have also noticed that today, I will remove that...

PR #1236, opened 2019-12-23
Combination chart point padding width, group width config

harellevy:


projects/ui-framework/bob-charts/src/charts/combinations/bar-line-chart/bar-line-chart.component.ts
```diff
@@ -49,12 +58,13 @@ export class BarLineChartComponent extends ChartCore implements OnChanges {
...
-
+        plotOptions: {
+          column: this.stacked
+              ? this.stackedPercent
+                ? {...COLUMN_BAR_WIDTH, stacking: 'percent'}
+                : {...COLUMN_BAR_WIDTH, stacking: 'normal'}
+              : {...MULTI_COLUMN_WIDTH, stacking: undefined}
+          }
       },
       { ...this.extraOptions }
```
galloween: you probably dont need the spread since you are using merge... and Object.assign would be event better then lodash...

projects/ui-framework/bob-charts/src/charts/combinations/bar-line-chart/bar-line-chart.component.ts
```diff
@@ -34,6 +34,15 @@ export class BarLineChartComponent extends ChartCore implements OnChanges {
   }
 
   private updateChartOptions() {
```
galloween: Capital letters consts should better be placed outside of class (can be same file), if it’s local to this function, better to name in camelCase...

PR #1258, opened 2019-12-25
Dany rtmn/numeric input ticker

danyRtmn:
## Status 
**READY**

## Description
Ticker buttons added to InputComponent. 
Ticker buttons increment/decrement inputs value by step
Ticker buttons are only displayed if step is truthy && inputType is number
Input is limited by min and max
Pushing decrement ticker button on empty input always sets the value to min limit
Pushing increment ticker button on empty input sets the value to min if step is less than min, and sets max if step is bigger than max

## Todos
- [x] Tests

input-component
base-input-element
password-input-component

Closing
1193


projects/ui-framework/src/lib/form-elements/base-input-element.ts
```diff
@@ -59,22 +60,21 @@ export abstract class BaseInputElement extends BaseFormElement {
...
-      ) {
+      if ((this.min && parsed < this.min) || (this.max && parsed > this.max)) {
         this.writeValue(parsed < this.min ? this.min : this.max);
         this.transmitValue(this.value, {
           eventType: [InputEventType.onChange],
         });
       }
     }
-
     this.transmitValue(this.value, { eventType: [InputEventType.onBlur] });
```
galloween: lets call this.onInputBlur() here, to not have code duplication

projects/ui-framework/src/lib/form-elements/input/input.component.spec.ts
```diff
@@ -112,4 +114,80 @@ describe('InputComponent', () => {
...
+  });
+
+  describe('Ticker buttons', () => {
+   it('should render ticker buttons if the inputType is number and step greater than 0', () => {
+     component.inputType = InputTypes.number;
+     component.step = 5;
+     fixture.detectChanges();
+     expect(elementFromFixture(fixture, '.b-icon-chevron-up')).toBeTruthy();
+     expect(elementFromFixture(fixture, '.b-icon-chevron-down')).toBeTruthy();
+     expect(elementFromFixture(fixture, '.bfe-buttons-wrap')).toBeTruthy();
```
galloween: testing just the wrapper should be enough, because we know that the buttons are inside

projects/ui-framework/src/lib/form-elements/input/input.component.ts
```diff
@@ -65,6 +66,18 @@ export class InputComponent extends BaseInputElement implements AfterViewInit {
...
+    this.writeValue(parsedValue + this.step);
+    this.processValue(parsedValue + this.step);
+  }
+
+  onDecrement() {
+    const parsedValue = parseToNumber(this.input.nativeElement.value);
+    this.writeValue(parsedValue - this.step);
+    this.processValue(parsedValue - this.step);
+  }
+
```
galloween: cool!

galloween: one thing, lets siimplify a little bit here:

```   
    this.writeValue(parseToNumber(this.input.nativeElement.value) + this.step);
    this.processValue(this.value);
```
  (because writeValue sets this.value). its the same thing as before, but looks a little easier to understand the flow

projects/ui-framework/src/lib/form-elements/input/input.stories.ts
```diff
@@ -102,6 +105,7 @@ story.add(
         maxChars: number('maxChars', 30),
         min: number('min', 5),
         max: number('max', 30),
+        step: number('step', 3),
```
galloween: you also need to add the description of [step] input in the Notes section - similar to other inputs...

projects/ui-framework/src/lib/form-elements/input/input.component.ts
galloween: or this.writeValue(parsedValue + this.step) if you want to keep the const.... same thing

projects/ui-framework/src/lib/form-elements/input/input.component.scss
```diff
@@ -1,3 +1,17 @@
...
+  border: none;
+}
+.b-icon-chevron-up {
+  border-bottom: 1px solid lightgrey;
+}
+
+.bfe-buttons-wrap {
+  background-color: white;
+  border-left: 1px solid lightgrey;
+}
```
galloween: still doesnt look like design! and also let's get rid from the blue outline when the button is clicked/focused

image

image

projects/ui-framework/src/lib/form-elements/input/input.component.scss
```diff
@@ -1,3 +1,29 @@
...
+.b-icon:before{
+  color: #aaa;
+}
+.b-icon-chevron-up {
+  border-bottom: 1px solid lightgrey;
+  border-top-right-radius: 4px;
+}
+.b-icon-chevron-down {
+  border-bottom-right-radius: 4px;
+}
```
galloween: Cool! looks much better. There are a few other tweaks we can apply to finalise it:

- remove `color: #aaa` from icon:before and `border: 1px solid #ccc` from .bfe-wrap 

- to make the icon lighter, add class 'b-icon-normal' to the buttons

- add 'type="button"' to the buttons (`<button type="button"....`) (buttons should always have type according to html5 spec. the other type is "submit" - used in forms)

- use `$input-border-color` variable instead of lightgrey (`border-left: 1px solid $input-border-color;`)   and `$border-radius` variable instead of 4px

- add the following rules to .b-icon:before: `height: ($input-height / 2);` `line-height: ($input-height / 2);` (to make the buttons height to be half of the input), remove `width: 100%` and `height: auto`

- to emulate the border between buttons that is exactly in the center, you can add these rules to `.bfe-buttons-wrap` :
```
    background-image: linear-gradient($input-border-color, $input-border-color);
    background-size: 100% 1px;
    background-position: center;
```
with this method you can remove `border-left` from the button

- move the buttons styles  inside `.bfe-buttons-wrap` (to make the selectors more precise, and not affect other b-icons we might have in the tempate). Like this: 

```
.bfe-buttons-wrap {
  border-left: 1px solid lightgrey;
  
  .b-icon {
    ...

    &:focus {
      outline: 0;
    }
  }

  .b-icon-chevron-up {
    ...
  }
  .b-icon-chevron-down {
    ...
  }
}
```

projects/ui-framework/src/lib/form-elements/input/input.stories.ts
```diff
@@ -70,6 +72,7 @@ const note = `
    input, label text (if present) will be used as placeholder
   [minChars] | number | minimum length
   [maxChars] | number | maximum length
+  [step] | number | (required for input to render ticker buttons) defines the ticker step
```
galloween: better comment would be:  `Step value for number input step buttons. Buttons will not be displayed, if step value is not provided.`

projects/ui-framework/src/lib/form-elements/input/input.stories.ts
```diff
@@ -61,7 +63,7 @@ const note = `
   #### Properties
   Name | Type | Description
   --- | --- | ---
-  [inputType] | InputTypes | type of input field
+  [inputType] | InputTypes | type of input field (inputType number is required to render ticker buttons)
```
galloween: only number inputs are supposed to have the buttons, so this addition is not necessary

projects/ui-framework/src/lib/form-elements/input/input.stories.ts
```diff
@@ -92,7 +95,8 @@ story.add(
       template: storyTemplate,
       props: {
         inputEvents: action('Input event'),
-        inputType: select('inputType', inputTypes, InputTypes.text),
+
+        inputType: select('inputType', inputTypes, InputTypes.number),
```
galloween: let's leave it with `InputTypes.text` as default

projects/ui-framework/src/lib/form-elements/input/input.component.scss
```diff
@@ -1,3 +1,33 @@
 @import '../../style/common-imports';
 
 @include bFormElement();
+
+.bfe-wrap{
+  border: 1px solid $input-border-color;
```
galloween: need to remove this, this element does not need any changes

projects/ui-framework/src/lib/form-elements/input/input.component.scss
```diff
@@ -1,3 +1,33 @@
...
+    }
+  }
+  .b-icon-chevron-up {
+    border-bottom: 1px solid $input-border-color;
+    border-top-right-radius: $border-radius;
+  }
+  .b-icon-chevron-down {
+    border-bottom-right-radius: $border-radius;
+  }
+}
```
galloween: cool!

projects/ui-framework/src/lib/form-elements/input/input.component.scss
```diff
@@ -1,3 +1,30 @@
...
+    &:focus{
+      outline: 0;
+    }
+    &:before{
+      height: ($input-height / 2);
+      line-height: ($input-height / 2);
+    }
+  }
+  .b-icon-chevron-up {
+    border-bottom: 1px solid $input-border-color;
```
galloween: and also remove the border bottom - because we use the background image for the border now...

danyRtmn:
> forgot to mention line 24 in css - we dont need the border-bottom anymore (using background image now)

yes, also forgot

PR #1301, opened 2020-01-07
Bump AgGrid version to 22.1.1 to enable new APIs

ishaiborovoy:


projects/ui-framework/bob-table/src/table/table/table.component.spec.ts
```diff
@@ -208,6 +208,15 @@ describe('TableComponent', () => {
             component.gridOptions.columnApi.autoSizeAllColumns
           ).toHaveBeenCalled();
         }));
+        it('should call autoSizeAllColumns when onGridReady is triggered', fakeAsync(() => {
```
omrihecht: should not

PR #1317, opened 2020-01-14
Pavel/vertical breadcrumbs

galloween:
breaking changes!
- buttons now not part of breadcrumbs
- interface change
- activeIndex input removed

here is the related bob2 PR: https://github.com/hibobio/hibob-web-client/pull/2429


galloween:
> The code looks good, let's merge tomorow when you are here, and fix all the breaking changes immidiately after

this is the related bob2 PR: https://github.com/hibobio/hibob-web-client/pull/2429
though it fails tests due to html changes

PR #1330, opened 2020-01-19
upgrade cells in grid

tomeraronovsky:
- wrap circle-icons component with actions-wrapper component to add actions capability
- new actionType value, when actionType = 'asMenu' the actions will be displayed in menu, when actionType = 'asInlineButton' the action will display as button

### DIFF
--- a/projects/ui-framework/bob-table/src/public_api.ts
+++ b/projects/ui-framework/bob-table/src/public_api.ts
@@ -24,4 +24,3 @@ export {
 export {
   CircleIconAndLabelParams
 } from './table/table-cell-components/circle-icon-and-label/circle-icon-and-label.interface';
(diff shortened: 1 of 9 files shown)


tomeraronovsky:
@omrihecht yes, there is situation that we have one action but they want it in menu, so this rule not enough..

omrihecht:
can you provide an example?

projects/ui-framework/bob-table/src/public_api.ts
```diff
@@ -24,3 +24,4 @@ export {
 export {
   CircleIconAndLabelParams
 } from './table/table-cell-components/circle-icon-and-label/circle-icon-and-label.interface';
+export { ActionsType } from './table/table-cell-components/table-actions-wrapper/table-actions-wrapper.enum';
```
omrihecht: you can probably remove this, it might throw an error since it does not exist anymore

PR #1336, opened 2020-01-26
rte overflow fix

benbaler:
closes https://github.com/hibobio/hibob/issues/32386

### DIFF
--- a/projects/ui-framework/bob-rte/src/rte/rte.component.scss
+++ b/projects/ui-framework/bob-rte/src/rte/rte.component.scss
@@ -15,7 +15,6 @@
 
 .bfe-wrap {
   position: relative;
-  z-index: 0;
   background-color: white;
   border: 1px solid $input-border-color;
   border-radius: $border-radius;


projects/ui-framework/bob-rte/src/rte/rte.component.scss
```diff
@@ -15,7 +15,6 @@
 
 .bfe-wrap {
   position: relative;
-  z-index: 0;
```
omrihecht: i'm not sure if this is the place...
this is the css of every form-element in bob
@galloween wdyt?

benbaler: @omrihecht its only for rte

galloween: @omrihecht @benbaler 
no, this is scoped to RTE, but
this change will re-introduce this bug: https://github.com/hibobio/hibob/issues/32155
different solution is needed, I will take a look

galloween:
closing this, fix in this PR: https://github.com/hibobio/bob-style/pull/1338 (problem is not rte, but tabs)

PR #1362, opened 2020-02-03
Pavel/side menu

galloween:
new side menu that supports avatars


projects/ui-framework/src/lib/navigation/menu/menu.component.ts
```diff
@@ -52,8 +64,8 @@ export class MenuComponent implements OnChanges {
   }
 
   onCloseMenu(): void {
-    if (this.closeMenu.observers.length > 0) {
-      this.closeMenu.emit(this.id || null);
-    }
+    // if (this.closeMenu.observers.length > 0) {
+    this.closeMenu.emit(this.id || null);
+    // }
```
doroncy: what is this comment out code? remove it

PR #1363, opened 2020-02-03
Pavel/side menu

galloween:
side menu keyboard control


projects/ui-framework/src/lib/services/html/dom-helpers.service.ts
```diff
@@ -215,6 +215,50 @@ export class DOMhelpers {
...
+  }
+
+  // TODO: Add Test
+  public getPrevSibling(
+    element: HTMLElement,
+    selector: string = null
+  ): HTMLElement {
+    return this.getSibling(element, selector, 'prev');
+  }
+
```
doroncy: great stuff, looks very usefull. need to add tests

projects/ui-framework/src/lib/navigation/side-menu/side-menu.component.html
```diff
@@ -7,19 +7,21 @@ <h4 class="header-title">{{ headerLabel }}</h4>
 
 <ul class="menu-options"
     (click.outside-zone)="onClick($event)"
+    (keydown.outside-zone)="onKey($event)"
     (focusin.outside-zone)="onFocus($event)"
     (focusout.outside-zone)="onBlur($event)">
```
doroncy: i like those output events. what are those? ```keydown.outside-zone```

galloween: Event Manager plugins. remember, we discussed them onece, maybe half a year ago ))
this one is supposed not to trigger global change detection.
but after we included the rxjs zone patch file, it doesnt really work... Im still using it in hope some day we will fix or rewrite the plugin. at the moment it doesnt make much difference

projects/ui-framework/src/lib/services/html/dom-helpers.service.ts
galloween: yes. then one day (ideally) we should migrate the whole service to use Render10 (or whatever iteration it will be at that point)...

PR #1376, opened 2020-02-09
add accent support to tables

omrihecht:
add accent support to tables


projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -194,6 +194,7 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
       onCellClicked(event: CellClickedEvent): void {
         that.cellClicked.emit(event);
       },
+      accentedSort: true
```
galloween: genius!

PR #1391, opened 2020-02-13
ignore chrome test files

pazt:


.gitignore
```diff
@@ -38,3 +38,6 @@ testem.log
 # System Files
 .DS_Store
 Thumbs.db
+
+# Chore output files
+*.!opt!google!chrome!chrome
```
galloween: need to escape the !

PR #1397, opened 2020-02-16
min height for regular size card-employee

simonScheff:
https://github.com/hibobio/hibob/issues/32874


projects/ui-framework/src/lib/cards/card-employee/card-employee.component.scss
```diff
@@ -20,6 +20,10 @@
     min-height: 200px;
     background-size: 100% 60px;
   }
+
+  &[data-type='regular'] {
+    min-height: 300px;
```
galloween: why? whats wrong with 280 (as it was before) ?

PR #1407, opened 2020-02-24
Mobile Service/get mobile os

oridanus:
to solve this - https://app.asana.com/0/1157695988834378/1160484435514194/f

### DIFF
--- a/projects/ui-framework/src/lib/services/utils/mobile.service.ts
+++ b/projects/ui-framework/src/lib/services/utils/mobile.service.ts
@@ -88,7 +88,7 @@ export class MobileService {
 
     // iOS detection from: http://stackoverflow.com/a/9039885/177710
     // @ts-ignore
-    if (/iPad|iPhone|iPod/.test(userAgent) && !window.MSStream) {
+    if (/ipad|iphone|ipod/i.test(userAgent) && !window.MSStream) {
       return MobileOS.iOS;
     }
 
--- a/projects/ui-framework/src/lib/services/utils/mobile.service.ts
+++ b/projects/ui-framework/src/lib/services/utils/mobile.service.ts
@@ -21,6 +21,7 @@ export interface MediaEvent {
   matchDesktop: boolean;
   isTouchDevice: boolean;
   isMobileBrowser: boolean;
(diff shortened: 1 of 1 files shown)


projects/ui-framework/src/lib/services/utils/mobile.service.ts
```diff
@@ -66,9 +74,29 @@ export class MobileService {
     return this.isMobBrowser;
   }
 
+  private getMobileOperatingSystem(): MobileOS {
```
galloween: what does it achieve? we can already detect touch devices and mobile browsers. why do we need to detect OS ?

projects/ui-framework/src/lib/services/utils/mobile.service.ts
```diff
@@ -32,12 +38,14 @@ export class MobileService {
...
   isMobBrowser: boolean;
   isTouchDevice: boolean;
+  mobileOS: MobileOS;
 
   constructor(
     private windowRef: WindowRef,
     private utilsService: UtilsService
   ) {
     this.isMobBrowser = this.checkForMobileBrowser();
+    this.mobileOS = this.getMobileOperatingSystem();
```
galloween: lets add it to getMediaData() method as well

oridanus: added

projects/ui-framework/src/lib/services/utils/mobile.service.ts
```diff
@@ -66,9 +74,29 @@ export class MobileService {
...
+      return MobileOS.windowsPhone;
+    }
+
+    if (/android/i.test(userAgent)) {
+      return MobileOS.android;
+    }
+
+    // iOS detection from: http://stackoverflow.com/a/9039885/177710
+    // @ts-ignore
+    if (/iPad|iPhone|iPod/.test(userAgent) && !window.MSStream) {
```
galloween: safer to test for lowercase + /i modifier (`/ipad|iphone|ipod/i`)

oridanus: fixed thanks!

PR #1449, opened 2020-03-18
Card table sortable

PovarovDenis:
**WORK IN PROGRESS**

New sortable table component based on the existing card-template component


projects/ui-framework/src/lib/table/card-table-sortable/card-table-sortable/card-table-sortable.component.scss
```diff
@@ -0,0 +1,75 @@
+@import '../../../style/common-imports';
```
galloween: no need to duplicate file, just import the original card table scss in the component styleUrls meta

projects/ui-framework/src/lib/table/card-table-sortable/card-table-sortable/card-table-sortable.component.ts
```diff
@@ -0,0 +1,43 @@
...
+//         icon: Icons.drag_alt,
+//         color: IconColor.dark,
+//         size: IconSize.small
+//       }
+//     }
+
+@Component({
+  selector: 'b-card-table-sortable',
+  templateUrl: './card-table-sortable.component.html',
+  styleUrls: ['./card-table-sortable.component.scss']
```
galloween: import the original card-table.component.scss here

projects/ui-framework/src/lib/table/card-table/card-table/card-table.component.scss
```diff
@@ -1,5 +1,5 @@
 @import '../../../style/common-imports';
-@import '../table-card-cell/table-card-cell-mixin';
+@import '../../table-card-cell/table-card-cell/table-card-cell-mixin';
```
galloween: this doesnt look right. the paths in table-card forlder did not change

projects/ui-framework/src/lib/table/card-table/card-table/card-table.component.ts
```diff
@@ -19,7 +19,7 @@ import {
   CardTableCellClickEvent,
 } from '../card-table.interface';
 import { CellWidthsService } from '../cell-widths-service/cell-widths.service';
-import { TableCardComponent } from './../table-card/table-card.component';
+import { TableCardComponent } from './../../table-card/table-card/table-card.component';
```
galloween: dont think this is needed

projects/ui-framework/src/lib/table/table-card-cell/table-card-cell.module.ts
```diff
@@ -0,0 +1,18 @@
+import { NgModule } from '@angular/core';
```
galloween: no need for this module

projects/ui-framework/src/lib/table/table-card-cell/table-card-cell/table-card-cell.component.ts
```diff
@@ -3,7 +3,7 @@ import { Component, Input } from '@angular/core';
 import {
   CardTableCellData,
   CardTableCellDataType
-} from '../card-table.interface';
+} from '../../card-table/card-table.interface';
```
galloween: put it back :)

projects/ui-framework/src/lib/table/table-card/table-card.module.ts
```diff
@@ -0,0 +1,16 @@
+import { NgModule } from '@angular/core';
```
galloween: no need for this module

projects/ui-framework/src/lib/table/table-card/table-card/table-card.component.html
```diff
@@ -1,3 +1,4 @@
+<ng-content></ng-content>
```
galloween: card table content is generated from  CardTableCellData[][] data. there is no way to pass ng-content here

projects/ui-framework/src/lib/table/table-card/table-card/table-card.component.ts
```diff
@@ -5,7 +5,7 @@ import {
   CardTableCellData,
   cardTableAllowedCellStyles,
   CardTableCellClickEvent
-} from '../card-table.interface';
+} from '../../card-table/card-table.interface';
```
galloween: revert

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.html
```diff
@@ -0,0 +1,49 @@
...
+      [row]="row"
+      [rowIndex]="index"
+      [meta]="meta"
+      [style]="cellsStyle"
+      [ngClass]="{'clickable': cellClicked.observers.length > 0 || rowClicked.observers.length > 0}"
+      (click)="onRowClicked(row, index)"
+      (cellClicked)="onCellClicked($event)"
+      role="row">
+    <div class="drag-handle-container" *ngIf="useDragHandle" cdkDragHandle>
+      <b-square-button [type]="buttons.tertiary"
```
galloween: lets put ```class="drag-handle-container" *ngIf="useDragHandle" cdkDragHandle``` directly on b-square-button

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.html
```diff
@@ -0,0 +1,49 @@
+<header role="rowgroup">
+  <ul class="card-table-header"
+      role="row">
+    <div class="drag-handle-container" *ngIf="useDragHandle"></div>
```
galloween: dont like it:
- u cant have div inside ul
- generally unnecessary element, maybe better solution is to add has-drag-handle (or similar) class to the host (see comment below) and then add a +50px padding-left to the card-table-header

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.html
galloween: also no need for class, you can use the cdkDragHandle attribute

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
```diff
@@ -0,0 +1,24 @@
...
+  opacity: 0;
+}
+.cdk-drag-animating {
+  transition: transform 200ms cubic-bezier(0, 0, 0.2, 1) !important;
+}
+.card-table-body.cdk-drop-list-dragging li:not(.cdk-drag-placeholder) {
+  transition: transform 200ms cubic-bezier(0, 0, 0.2, 1);
+}
+
+.drag-handle-container {
```
galloween: you can use ```[cdkDragHandle]```
or keep the class, but change to .drag-handle (because we remove the unnecessary wrapper)

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.html
galloween: also you are not using the button functionality here (nothing happens when you press it)
so its better to use <b-icon> here instead

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
```
galloween: you can achieve the same with 1 less line:

display: flex;
align-items: center;

BUT! if you switch to b-icon, you wont need it at all, because it has it inside already.

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
```diff
@@ -0,0 +1,24 @@
...
+  padding: 0 10px;
+  margin: 0;
+  width: 50px;
+
+  ::ng-deep {
+    button {
+      cursor: move !important;
```
galloween: move this to the element itself ([cdkDragHandle])

galloween: ... and u wont need important

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  selector: 'b-card-table-sortable',
+  templateUrl: './card-table-sortable.component.html',
+  styleUrls: [
+    './card-table-sortable.component.scss',
+    './../card-table/card-table.component.scss',
+  ],
+})
+export class CardTableSortableComponent extends CardTableComponent
+  implements OnInit {
+  @Input() useDragHandle = false;
```
galloween: you can bind it to an attribute on the host:

```
@HostBinding('attr.data-has-drag-handle') @Input() useDragHandle = false;
```

and then use it in css to add padding to header:

```
:host[data-has-drag-handle="true"] .card-table-header {
 padding-left: 50px // + padding-left that was there before, if any
}
```

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  public readonly iconSize = IconSize;
+  public readonly iconColor = IconColor;
+  public readonly buttons = ButtonType;
+
+  ngOnInit(): void {
+    this.setCellsStyle();
+  }
+
+  onDrop(event: CdkDragDrop<CardTableCellData[][]>) {
+    moveItemInArray(this.table, event.previousIndex, event.currentIndex);
```
galloween: wow cool, didnt know they had this

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.html
```diff
@@ -0,0 +1,46 @@
...
+        }"
+        role="columnheader">
+      <span class="text">{{ cell.name }}</span>
+    </li>
+  </ul>
+</header>
+
+<ul class="card-table-body"
+    cdkDropList
+    [cdkDropListData]="table"
```
galloween: you are not using this data

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
```diff
@@ -0,0 +1,16 @@
...
+  opacity: 0;
+}
+.cdk-drag-animating {
+  transition: transform 200ms cubic-bezier(0, 0, 0.2, 1) !important;
+}
+.card-table-body.cdk-drop-list-dragging li:not(.cdk-drag-placeholder) {
+  transition: transform 200ms cubic-bezier(0, 0, 0.2, 1);
+}
+:host[data-has-drag-handle="true"] .card-table-header .table-header-cell:first-of-type {
+  margin-left: 50px;
```
galloween: better to use padding-left 60px (it had 10 before and now + 50)
this way the spacing is controlled in 1 property, not 2

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    super(widthsService);
+  }
+
+  public readonly icons = Icons;
+  public readonly iconSize = IconSize;
+  public readonly iconColor = IconColor;
+
+  ngOnInit(): void {
+    this.setCellsStyle();
+  }
```
galloween: this already exists in the parent class, so no need to duplicate

projects/ui-framework/src/lib/table/card-table/card-table/card-table.component.ts
```diff
@@ -65,6 +65,7 @@ export class CardTableComponent implements OnInit {
   }
 
   onRowClicked(row: CardTableCellData[], index: number): void {
+    console.log('click', row, index);
```
galloween: dont forget to remove log

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.ts
PovarovDenis: uf yeah, forgot to remove, previously there was my logic

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
PovarovDenis: I tried, but using padding-left on .card-table-header ruins somehow the layout of the header columns..
Screenshot from 2020-03-19 11-27-33

galloween: not on the header, on the first .table-header-cell, just padding instead of margin

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.ts
```diff
@@ -0,0 +1,48 @@
...
+@Component({
+  selector: 'b-card-table-sortable',
+  templateUrl: './card-table-sortable.component.html',
+  styleUrls: [
+    './card-table-sortable.component.scss',
+    './../card-table/card-table.component.scss',
+  ],
+})
+export class CardTableSortableComponent extends CardTableComponent
+  implements OnInit {
```
galloween: you can also remove the implements OnInit ;)

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
PovarovDenis: I tried that also but it doesn't work also
Screenshot from 2020-03-19 12-24-05
actually it was my initial attempt to use padding, but it doesn't work and I used that weird solution with div :)

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
```diff
@@ -0,0 +1,16 @@
...
+}
+[cdkDragHandle] {
+  width: 50px;
+  cursor: move;
+}
```
galloween: also you need to make the rows height smaller:
```
:host ::ng-deep {
  [b-table-card], .table-card-empty {
    min-height: 55px;
  }
}
```

projects/ui-framework/src/lib/table/card-table/card-table-sortable/card-table-sortable.component.scss
galloween: it doesnt work but for different reason. the cells in the row have width in %, the button is an extra element now that is not expected there, so it pushes other elements.
i think the solution here is to:
- add padding to both .card-table-header and .card-table-body (not first cells)
- position the drag handle icon absolutely in the empty space created by the padding - that way it should not interfere with the width calculation

galloween: something like this

```
:host[data-has-drag-handle="true"]{
  .card-table-header, [b-table-card] {
    padding-left: 60px;
  }

  [b-table-card] {
    position: relative;
  }

  [cdkDragHandle] {
    position: absolute;
    left: 10px;
    top: 0;
    bottom: 0;
  }
}
```

galloween: you solution with margin also works actually, we could leave it like that, but im not sure leaving that extra element (drag handle) and not considering it for cell width calculation will behave predictably in the future.
i think its safer to remove it from the flow and position absolutely.

PR #1509, opened 2020-05-06
directory selectFilteredOnly

omrihecht:
https://app.asana.com/0/1159956917255868/1174481783099277/f

### DIFF
--- a/projects/ui-framework/bob-table/src/table/table.stories.ts
+++ b/projects/ui-framework/bob-table/src/table/table.stories.ts
@@ -10,7 +10,7 @@ import {
 } from '@storybook/addon-knobs/angular';
 import { storiesOf } from '@storybook/angular';
 import { AgGridModule } from 'ag-grid-angular';
-import { AvatarModule, ComponentGroupType } from 'bob-style';
+import { AvatarModule, ComponentGroupType, SearchModule } from 'bob-style';
 import { values } from 'lodash';
 import { StoryBookLayoutModule } from '../../../src/lib/story-book-layout/story-book-layout.module';
 import { ActionsCellComponent } from './table-cell-components/actions-cell/actions-cell.component';
@@ -47,7 +47,7 @@ const story = storiesOf(ComponentGroupType.Tables, module).addDecorator(
 );
 
 const template = `
-<b-table
+<b-table #table
   [type]="type"
   [rowData]="rowData"
   [columnDefs]="columnDefs"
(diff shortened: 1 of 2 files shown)


projects/ui-framework/bob-table/src/table/table.stories.ts
```diff
@@ -57,7 +59,7 @@ const template = `
   [shouldAutoSizeColumns]="shouldAutoSizeColumns"
   (rowClicked)="rowClicked($event)"
   (cellClicked)="cellClicked($event)"
-  (selectionChanged)="selectionChanged($event)"
+  (selectionChanged)="selectionChanged($event); onSelectionChanged($event)"
```
doroncy, line 62: whats this ?

omrihecht: i added search + filter in the story of b-table, it's for that
image

projects/ui-framework/bob-table/src/table/table.stories.ts
```diff
@@ -217,6 +219,11 @@ function tableStoryFactory({
     sortChanged: action('Sort changed'),
     columnsOrderChanged: action('Column order changed'),
     columnRemoved: action('Column remove button clicked'),
+    totalSelected: 0,
+    onSearchChange: (str, table) => table.gridOptions.api.setQuickFilter(str),
+    onSelectionChanged: function($event) {
+      this.totalSelected = $event.length;
```
galloween, line 225: cool! `this` works! 
you could also do it in the onSearchChange method

projects/ui-framework/bob-table/src/table/table.stories.ts
```diff
@@ -47,7 +47,9 @@ const story = storiesOf(ComponentGroupType.Tables, module).addDecorator(
 );
 
 const template = `
-<b-table
+<b-search style="display: block; width: 250px; margin-bottom: 8px;" (searchChange)="onSearchChange($event, table)"></b-search>
+<div style="margin-bottom: 16px;">total selected: {{totalSelected}}</div>
+<b-table #table
```
galloween, line 52: this also goes to Notes - check if it looks ok? sometimes if there is too much not related stuff, i do a separate 'template', just for the notes (without the extras)...

projects/ui-framework/bob-table/src/table/table.stories.ts
omrihecht: i need to update the selected number in the selectChange event

galloween: ah, 'selected' is not same as 'search results length'

omrihecht: Btw, this works because i used function instead of arrow function

PR #1519, opened 2020-05-12
added headerClass to colDef interface

doroncy:
1. added headerClass and autoHeight to columnDef interface
2. added bg-grey-100 common class for easy use of common bg classes.

### DIFF
--- a/projects/ui-framework/bob-table/src/table/table/table.interface.ts
+++ b/projects/ui-framework/bob-table/src/table/table/table.interface.ts
@@ -25,6 +25,7 @@ export interface ColumnDef {
   flex?: number;
   colSpan?: Function;
   autoHeight?: boolean;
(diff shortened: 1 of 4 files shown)


projects/ui-framework/bob-table/src/table/table/table.interface.ts
```diff
@@ -23,6 +24,7 @@ export interface ColumnDef {
   getQuickFilterText?: Function;
   flex?: number;
   colSpan?: Function;
+  autoHeight?: boolean;
```
galloween, line 27: omri suggested to also add [key:string]: any so that more ag-grid stuff cant be passed without the need to update the interface

doroncy: yes, will add this.

PR #1525, opened 2020-05-19
Pavel/translations support

galloween:
**Note**
need to remove [translation] input related code from bob2 (for editable list / tree-list)


projects/ui-framework/src/lib/i18n/i18n.module.ts
```diff
@@ -13,11 +13,12 @@ export class StorybookTranslateModule {
     translateService: TranslateService,
     private httpClient: HttpClient
   ) {
-    console.log('Translations: ', translateService.translations);
     this.httpClient.get(`${translationFilePrefix}en.json`).subscribe((data) => {
       translateService.setTranslation('en', data, true);
       translateService.setDefaultLang('en');
       translateService.use('en');
+
+      console.log('Translations: ', translateService.translations);
```
doroncy, line 21: we dont need this console log

PR #1542, opened 2020-05-26
Pavel/formatNumber pipe

galloween:
- formatNumber pipe
- number input displays formatted number, when not in focus

image


projects/ui-framework/src/lib/services/filters/formatNumber.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+  pure: true,
+})
+export class FormatNumberPipe implements PipeTransform {
+  constructor() {}
+
+  private formatter = new Intl.NumberFormat('en', {
+    style: 'decimal',
+    useGrouping: true,
+    maximumFractionDigits: 3,
+  });
```
doroncy, line 15: This is amazing!! but it will be much better if we use the correct language and not always en

PR #1559, opened 2020-06-09
sortable-collapsible-sections

omrihecht:
closes https://app.asana.com/0/1159939249943132/1179574909506137/f

### DIFF
--- a/projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
+++ b/projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
@@ -1,0 +1,60 @@
+<ul class="sections"
+    [ngClass]="{'dragging': dragging}"
+    (cdkDropListDropped)="onDrop($event)"
+    cdkDropList>
+
+  <li *ngFor="let section of UISections; trackBy: trackId; let index=index"
+      class="draggable-section"
+      cdkDrag
+      cdkDragLockAxis="y"
+      cdkDragStartDelay="5"
+      cdkDragBoundary="sections"
+      (cdkDragStarted)="dragStart.emit(index); draggedSection=section"
+      (cdkDragEnded)="dragEnd.emit(index)">
+
+    <b-collapsible-section class="section"
+                           [collapsible]="true"
+                           [expanded]="section.expanded || false"
+                           [title]="section.title"
+                           [description]="section.description"
+                           [options]="section.collapsibleOptions"
+                           (closed)="onClosed(index); section.expanded=false"
+                           (opened)="opened.emit(index); dragging=false"
+                           (openedFirst)="onContentLoad(index); openedFirst.emit(index)">
+
+      <ng-container header>
+        <div cdkDragHandle
+             class="drag-handler"
+             (click)="$event.stopPropagation()"
+             (mousedown)="onDragHandlerMouseDown()">
+        </div>
+        <ng-container *ngTemplateOutlet="collapsibleHeader?.tpl; context: { data: section.headerData, index: index }">
+        </ng-container>
+      </ng-container>
+
+      <ng-container *ngIf="contentLoadedMap.get(index)">
+        <ng-container *ngTemplateOutlet="collapsibleContent?.tpl; context: { data: section.contentData, index: index }">
+        </ng-container>
+      </ng-container>
+
+    </b-collapsible-section>
+
+  </li>
+
+  <ng-template cdkDragPreview
+               [data]="draggedSection">
+    <b-collapsible-section class="section"
+                           [collapsible]="true"
+                           [expanded]="false"
+                           [title]="draggedSection.title"
+                           [description]="draggedSection.description">
+
+      <ng-container header>
+        <ng-container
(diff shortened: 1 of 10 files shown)


projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
```diff
@@ -0,0 +1,59 @@
...
+                             [expanded]="false"
+                             [title]="section.title"
+                             [description]="section.description">
+
+        <ng-container header>
+          <ng-container *ngTemplateOutlet="collapsibleHeader?.tpl; context: { data: section.headerData, index: index }">
+          </ng-container>
+        </ng-container>
+      </b-collapsible-section>
+    </ng-template>
```
galloween, line 55: - i think you need to remove the matchSize, it might be asking for trouble
- if you want to optimize it a bit, you can put this template outside the li and ngFor (so its 1 template and not 10) -
cdkDragPreview directive has [data] input that it uses for context, but i have no idea how to use it - need to try :)

omrihecht: The match size means that the preview header will be the same size as the original, this works well, if you remove it, the header will take the size of the content
i'll check out the preview data

projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.ts
```diff
@@ -0,0 +1,97 @@
...
+    private cdr: ChangeDetectorRef
+  ) {
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.sections && changes.sections.currentValue) {
+      this.UISections = cloneArray(this.sections);
+      this.contentLoadedMap = new Map(
+        this.UISections.map((section, index) => [index, section.expanded])
+      );
```
galloween, line 58: just a note, you could use simple array for the expanded state:

// init - no need to init!
this.contentLoaded: boolean[] = []

// set expanded :  
contentLoaded[index] = true

// check expanded
if contentLoaded[index] <- surprise!

projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.module.ts
```diff
@@ -0,0 +1,33 @@
...
+    DragDropModule
+  ],
+  exports: [
+    SortableCollapsibleSectionsComponent,
+    CollapsibleContentDirective,
+    CollapsibleHeaderDirective,
+  ],
+  providers: [
+    DOMhelpers,
+    EventManagerPlugins[0]
```
galloween, line 29: dont think you use these

projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.scss
```diff
@@ -0,0 +1,73 @@
...
+    @include b-icon(
+      $elem: '&:after',
+      $icon: 'drag-alt',
+      $size: 'medium',
+      $color: 'light',
+      $hover: true
+    ) {
+      order: -1;
+      margin-right: times8(1);
+      margin-left: - times8(2);
```
galloween, line 58: this syntax hurts my eye a little bit (specifically the space there)... 
i would put `times8(-2)` or `times8(2) * -1`

projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
galloween: weird. i never used the matchsize in my stuff. cant imagine why it would be different size if its the same template

projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.ts
```diff
@@ -0,0 +1,98 @@
...
+@Component({
+  selector: 'b-sortable-collapsible-sections',
+  templateUrl: './sortable-collapsible-sections.component.html',
+  styleUrls: ['./sortable-collapsible-sections.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class SortableCollapsibleSectionsComponent implements OnChanges {
+
+  @ContentChild(CollapsibleHeaderDirective, { static: true }) collapsibleHeader !: CollapsibleHeaderDirective;
+  @ContentChild(CollapsibleContentDirective, { static: true }) collapsibleContent !: CollapsibleContentDirective;
```
omrihecht, line 30: @danidin , this `!:` is something i copied from your implementation in the chain select, i don't know what it means, can you please explain

galloween: some (newer) typescript concept:
definite assignment assertions

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions

PR #1565, opened 2020-06-14
adding col config

ronyrosenberg:
HOLD - missing manual sanity with the version in web client
Second round to get feedback.


projects/app-test/src/app/test-table/test-table.component.ts
```diff
@@ -40,17 +42,28 @@ export class TestTableComponent implements OnInit {
       }
     ];
     this.selectValues = this.fields;
-
+    this.columnDefConfig = {
+      columnDef: this.testTableService.getColumnDef(this.fields),
+      orderStrategy: ColumnOrderStrategy.Preserve,
+    };
```
omrihecht, line 48: i think this should go to onChanges as well, if you don't have this input, but only columnDef input, we should assign it, so internally we are always working on the same objects

projects/app-test/src/app/test-table/test-table.component.ts
```diff
@@ -40,17 +42,28 @@ export class TestTableComponent implements OnInit {
...
     this.updateTable();
   }
 
   onSelectChanged(listChange: ListChange) {
     this.fields = listChange.getSelectedIds() as string[];
+    this.columnDefConfig = {
+      columnDef: this.testTableService.getColumnDef(this.fields),
+      orderStrategy: ColumnOrderStrategy.AddToLast,
+    };
```
omrihecht, line 66: i am not sure this is relevant, isn't select the checkboxes?

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -140,7 +142,18 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
...
+      const columnDefs = (changes.columnDefConfig.currentValue as ColumnDefConfig).orderStrategy === ColumnOrderStrategy.AddToLast
+        ? this.tableUtilsService.getOrderedFields(this.gridColumnDefs, changes.columnDefConfig.currentValue.columnDef, )
+        : changes.columnDefConfig.currentValue.columnDef;
+
+      this.gridColumnDefs = this.tableUtilsService.getGridColumnDef(
+        columnDefs,
+        this.rowSelection
+      );
+    }
+
```
omrihecht, line 155: i would suggest to create columnDefConfig object if one is not provided, so we will be working on the same object

projects/ui-framework/bob-table/src/table/table/table.enum.ts
```diff
@@ -18,3 +18,8 @@ export enum PinDirection {
   Left = 'left',
   Right = 'right',
 }
+
+export enum ColumnOrderStrategy {
+  AddToLast = 'addToLast',
+  Preserve = 'preserve',
+}
```
omrihecht, line 25: you should export it in the public_api

projects/ui-framework/bob-table/src/table/table/table.interface.ts
```diff
@@ -52,3 +52,8 @@ export interface TableStyleConfig {
   disableRowHoverBgColor?: boolean;
   showColumnBorders?: boolean;
 }
+
+export interface ColumnDefConfig {
+  columnDef: ColumnDef[];
+  orderStrategy: ColumnOrderStrategy;
+}
```
omrihecht, line 59: you should export it in the public_api

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
```diff
@@ -83,4 +83,16 @@ export class TableUtilsService {
...
+    return columnDef.map((colDef) => colDef.field);
+  }
+
+  getOrderedFields(existingColumns: ColumnDef[], newColumns: ColumnDef[]): ColumnDef[] {
+    const newColsFiels = this.getColumnsField(newColumns);
+    const existingColsFiels = this.getColumnsField(existingColumns);
+    const fieldsToKeep = existingColumns.filter(column => newColsFiels.includes(column.field));
+    const fieldsToAdd = newColumns.filter(column => !existingColsFiels.includes(column.field));
+    return [...fieldsToKeep, ...fieldsToAdd];
+  }
```
omrihecht, line 97: lgtm, @galloween , if you can, please take a look, you are good with this kind of things...

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -66,6 +67,7 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
 
   @Input() rowData: any[] = [];
   @Input() columnDefs: ColumnDef[] = [];
+  @Input() columnDefConfig: ColumnDefConfig = { columnDef: [], orderStrategy: ColumnOrderStrategy.Preserve };
```
omrihecht, line 70: no need for default value here

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
ronyrosenberg: It was @omer-r 's code

projects/ui-framework/bob-table/src/table/table/table.interface.ts
ronyrosenberg: interface and enum are already exported
image

projects/ui-framework/bob-table/src/table/table/table.enum.ts
galloween: bob-table's public_api already exports * from this file, so it should cover it

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
galloween: if this is meant to combine 2 arrays without duplicates, then you can use this helper function (import {joinArrays} from 'bob-style'):

`return joinArrays(fieldsToKeep, fieldsToAdd)`

just for documentation sake, there are also: arrayIntersection(arr1,arr2) and arrayDifference(arr1,arr2)

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
```diff
@@ -83,4 +83,16 @@ export class TableUtilsService {
       })
       : null;
   }
+
+  private getColumnsField(columnDef: ColumnDef[]): string[] {
+    return columnDef.map((colDef) => colDef.field);
+  }
+
+  getOrderedFields(existingColumns: ColumnDef[], newColumns: ColumnDef[]): ColumnDef[] {
+    const newColsFiels = this.getColumnsField(newColumns);
```
omer-r, line 92: Typo 🧐

projects/ui-framework/bob-table/src/table/table/table.enum.ts
```diff
@@ -18,3 +18,8 @@ export enum PinDirection {
   Left = 'left',
   Right = 'right',
 }
+
+export enum ColumnOrderStrategy {
+  AddToLast = 'addToLast',
+  Preserve = 'preserve',
```
omer-r, line 24: Since the object is carrying all the fields (not just the additions) addToLast and preserve actually have a similar meaning...
Suggest to be more specific and use term like “appendNew” which is the js equivalent to “add last” and “reorder” which means the order of all columns may change

projects/ui-framework/bob-table/src/table/table/table.component.ts
omer-r: Strongly second that. You either get columnDef or columnDefConfig but after this check you continue with columnDefConfig which is the more generalised form

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
```diff
@@ -83,4 +83,24 @@ export class TableUtilsService {
...
+  getOrderedFields(
+    existingColumns: ColumnDef[],
+    newColumns: ColumnDef[],
+    columnsOrder: string[]): ColumnDef[] {
+    const newColsFields = this.getColumnsField(newColumns);
+    const sortedExistingCols = columnsOrder
+      ? existingColumns.sort((a, b) => {
+        return columnsOrder.indexOf(a.field) - columnsOrder.indexOf(b.field);
+      })
+      : existingColumns;
```
omrihecht, line 100: i think you can get the order of the columns they are currently sorted out in from the grid-api, checkout
`onSortChanged`...

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -140,9 +142,37 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
...
+    if (has(changes, 'columnDefConfig')) {
+      this.columnDefConfig = changes.columnDefConfig.currentValue;
+      const existingColumns = changes.columnDefConfig.previousValue
+        ? (changes.columnDefConfig.previousValue as ColumnDefConfig).columnDef
+        : this.columnDefs;
+      const columnDefs = this.columnDefConfig.orderStrategy === ColumnOrderStrategy.AppendNew
+        ? this.tableUtilsService.getOrderedFields(
+          existingColumns,
+          this.columnDefConfig.columnDef,
+          this.columns)
```
omrihecht, line 154: right.... so this.columns is string[] of the columns order right?

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -140,9 +142,37 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
...
+      this.columnDefConfig = {
+        columnDef: changes.columnDefs.currentValue,
+        orderStrategy: ColumnOrderStrategy.AppendNew,
+      };
+      const existingColumns = changes.columnDefs.previousValue ? changes.columnDefs.previousValue : this.columnDefs;
+      const columnDefs = this.tableUtilsService.getOrderedFields(
+        existingColumns,
+        this.columnDefConfig.columnDef,
+        this.columns
+      );
```
omrihecht, line 173: it's a bit of duplication, i think we can make this a bit prettier, wdyt?

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
```diff
@@ -83,4 +83,24 @@ export class TableUtilsService {
...
+    const existingColsFields = this.getColumnsField(sortedExistingCols);
+    const fieldsToKeep = existingColumns.filter(column => newColsFields.includes(column.field));
```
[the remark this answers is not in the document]
ronyrosenberg: because I'm sending back columnDefs and not paths (fields)

projects/ui-framework/bob-table/src/table/table/table.component.ts
ronyrosenberg: exactly, that's the order

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
ronyrosenberg: I tried. It's coming from an event triggered from the table that I don't have access to.

projects/ui-framework/bob-table/src/table/table/table.component.ts
ronyrosenberg: Its not exactly the same. Column def config has columnDef inside so the changes are not in the same level

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -140,12 +142,36 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
   }
 
   ngOnChanges(changes: SimpleChanges): void {
+    let updateColumns = false;
+    let previousColumnDefValue: ColumnDef[];
     if (has(changes, 'columnDefs')) {
+      updateColumns = true;
+      this.columnDefConfig = {
+        columnDef: changes.columnDefs.currentValue,
+        orderStrategy: ColumnOrderStrategy.AppendNew
```
omrihecht, line 151: this.columnDefConfig = {...columnDefConfigDefault, ...{columnDef: changes.columnDefs.currentValue}}

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.spec.ts
```diff
@@ -49,6 +50,45 @@ describe('TableUtilsService', () => {
...
+      {
+        headerName: 'Department',
+        field: 'department',
+        resizable: true,
+        sortable: true,
+        menuTabs: [],
+        cellClass: [],
+        cellStyle: {},
+      },
+    ];
```
omrihecht, line 91: let's keep this to minimum, you only need field

projects/ui-framework/bob-table/src/table/table/table.component.spec.ts
```diff
@@ -276,6 +286,80 @@ describe('TableComponent', () => {
         }));
       });
     });
+    fdescribe('colDefOrder', () => {
```
omrihecht, line 289: remove fdescribe

PR #1566, opened 2020-06-14
add recursiveFilter

omrihecht:


projects/ui-framework/src/lib/services/utils/functional-utils.ts
```diff
@@ -414,6 +414,16 @@ export const mapSplice = <K = any, V = any>(
   return deletedElementsMap;
 };
 
+export const recursiveFilter = (array: any[], childrenKey: string, fn: Function) => {
+  return array.reduce((acc, o) => {
+    const children = recursiveFilter(o[childrenKey] || [], childrenKey, fn);
```
galloween, line 419: i think this should go inside  if (fn(o))

PR #1576, opened 2020-06-15
add id to draggable-section

omrihecht:
to be used for instance to scroll to an element

### DIFF
--- a/projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
+++ b/projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
@@ -4,6 +4,7 @@
     cdkDropList>
 
   <li *ngFor="let section of UISections; trackBy: trackId; let index=index"
+      id="{{section.id}}"
       class="draggable-section"
       cdkDrag
       cdkDragLockAxis="y"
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/layout/sortable-collapsible-sections/sortable-collapsible-sections.component.html
```diff
@@ -4,6 +4,7 @@
     cdkDropList>
 
   <li *ngFor="let section of UISections; trackBy: trackId; let index=index"
+      id="{{section.id | json}}"
```
omrihecht, line 7: `| json` makes string, not sure what happens when you have `id="'0'"`

galloween: Not sure why you would put json pipe on a string or number. 
What’s wrong with id=“0” ? 
If you worry about them not being unique enough you can prefix it with something. id=“‘bcss-’+section.id”

omrihecht: i thought you could not set id=number

omrihecht: seems to be ok, though, removing | json

PR #1583, opened 2020-06-21
fix for expandable header content

omrihecht:
image

to this:

image


galloween:
or to this:

image

omrihecht:
so how do we solve it?

galloween:
> so how do we solve it?
> ::ng-deep inside the feature?

the [header] element (that should have `display: flex`) is in the feature template, so i think ng-deep will be not needed, but not 100% sure

PR #1588, opened 2020-06-25
Added toolTipSummary to table-actions-wrapper component

maximblack:
https://app.asana.com/0/1166845715193929/1181583085791737/f

### DIFF
--- a/projects/ui-framework/bob-table/src/table/table-cell-components/table-actions-wrapper/table-actions-wrapper.component.html
+++ b/projects/ui-framework/bob-table/src/table/table-cell-components/table-actions-wrapper/table-actions-wrapper.component.html
@@ -7,7 +7,7 @@
         [size]="iconSize.medium"
         [type]="iconType"
         [icon]="icon"
-        [toolTipSummary]="tooltipSummary">
+        [toolTipSummary]="toolTipSummary">
 </b-icon>
 
 <b-menu *ngIf="menuItems && menuItems.length > 1"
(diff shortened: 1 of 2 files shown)


projects/ui-framework/bob-table/src/table/table-cell-components/table-actions-wrapper/table-actions-wrapper.component.html
```diff
@@ -6,7 +6,8 @@
         [color]="iconColor.normal"
         [size]="iconSize.medium"
         [type]="iconType"
-        [icon]="icon">
+        [icon]="icon"
+        [toolTipSummary]="toolTipSummary">
```
galloween, line 10: better use `[matTooltip]="toolTipSummary"` instead, in this case.
(import MatTooltipModule)

projects/ui-framework/bob-table/src/table/table-cell-components/table-actions-wrapper/table-actions-wrapper.component.ts
```diff
@@ -10,6 +10,7 @@ export class TableActionsWrapperComponent {
   @Input() menuItems: MenuItem[];
   @Input() buttonType: ButtonType = ButtonType.primary;
   @Input() icon: Icons = null;
+  @Input() toolTipSummary: string;
```
galloween, line 13: maybe `iconTooltip` would be more precise

PR #1640, opened 2020-07-22
WIP - hide pager if only one page available

tomeraronovsky:
hide pager if only one page available >>
image


galloween:
This functionality was built in in the table, but was removed by request from @doroncy and @omrihecht

galloween:
this is now re-implemented

PR #1655, opened 2020-07-28
Some changes in `columnsChanged` and `columnsOrderChanged` events

PovarovDenis:
## b-table component
`columnsOrderChanged` event changes: added *eventName* property to the event
`columnsChanged` event changes: now event emits list of columns. Also event changes the internal property *columns* (to be consistent with the event name).

### DIFF
--- a/projects/ui-framework/bob-table/src/table/table/table.interface.ts
+++ b/projects/ui-framework/bob-table/src/table/table/table.interface.ts
@@ -1,5 +1,5 @@
 import {
-  ColumnOrderStrategy, ColumnsOrderChangedEventName,
+  ColumnOrderStrategy, TableEventName,
   PinDirection,
   SortDirections,
 } from './table.enum';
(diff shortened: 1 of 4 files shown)


projects/ui-framework/bob-table/src/table/table/table.interface.ts
```diff
@@ -50,6 +50,11 @@ export interface RowNodeDef {
 
 export interface ColumnsOrderChangedEvent {
   columns: string[];
+  eventName: ColumnsOrderChangedEventName;
+}
+
+export interface ColumnsChangedEvent {
+  columns: string[];
 }
```
galloween, line 58: - lets make eventName? optoinal (for backwards compatibility)

- even better:

```
export interface ColumnsChangedEvent {
  columns: string[];
}

export interface ColumnsOrderChangedEvent extends ColumnsChangedEvent {  
  eventName?: TableEventName;
}
```

PR #1695, opened 2020-08-18
Added outside param for chart tooltip

maximblack:
https://app.asana.com/0/1160257659441133/1188441932815889/f


galloween (approved):

galloween:
> What about adding the `outside: false` to the extra graph options we have in `KpisGraphService` it should override the default options and you can check it and see the change immediately

if you pass `tooltip: { outside: true }` in the `[extraOptions]`, it will override all the tooltip defaults, because its not deep-merging

PR #1728, opened 2020-09-17
Colors - add new colors palette

rinat-shemer:
https://app.asana.com/0/1159939249943132/1191725662159777/f


galloween (approved):

projects/ui-framework/src/lib/style/colorsPalette.enum.ts
```diff
@@ -0,0 +1,47 @@
+export enum ColorPelette {
```
galloween, line 1: - colorP**A**lette ?
- also lets move the file to `projects/ui-framework/src/lib` folder (no better location for now)
- also need to export * from '...colorsPalette.enum' in `projects/ui-framework/src/public_api.ts`

galloween (changes requested):

galloween (approved):

PR #1730, opened 2020-09-18
Maxdsr/add input for border left for collapsible section

MaxDsr:
Add the [showLeftSideBorder] input to wether display or not the colored border on the left for the `<b-collapsible-section>` component.

Screen Shot 2020-09-18 at 16 40 55

### DIFF
--- a/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
+++ b/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
@@ -2,7 +2,8 @@
          [attr.aria-disabled]="disabled || null"
          [attr.data-collapsible]="!!collapsible"
          [attr.data-expanded]="!collapsible ? null : !!expanded"
-         [attr.data-divided]="!!divided">
+         [attr.data-divided]="!!divided"
+         [attr.show-border]="!!options.indicatorColor">
 
   <header *ngIf="title || hasHeaderContent"
           class="bcp-header {{panelID}}"
@@ -11,7 +12,6 @@
           [attr.aria-controls]="(collapsible && panelID) || null"
           [attr.aria-expanded]="!collapsible ? null : !!expanded"
           [attr.data-icon-before]="!collapsible ? null : expanded ? 'chevron-down' : 'chevron-right'"
-          [class.show-border]="showLeftSideBorder"
           (click)="togglePanel()">
 
     <div *ngIf="title"
@@ -40,8 +40,7 @@
        [attr.id]="panelID"
        [attr.aria-labelledby]="panelID+'-header'"
        [attr.data-collapsible]="!!collapsible"
-       [attr.data-expanded]="!collapsible || startsExpanded ? null : !!expanded"
-       [class.show-border]="showLeftSideBorder">
+       [attr.data-expanded]="!collapsible || startsExpanded ? null : !!expanded">
 
     <div class="bcp-panel-content-wrap {{panelID}}"
          [attr.data-collapsible]="!!collapsible"
--- a/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
+++ b/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
@@ -11,6 +11,7 @@
           [attr.aria-controls]="(collapsible && panelID) || null"
           [attr.aria-expanded]="!collapsible ? null : !!expanded"
           [attr.data-icon-before]="!collapsible ? null : expanded ? 'chevron-down' : 'chevron-right'"
+          [class.show-border]="showLeftSideBorder"
           (click)="togglePanel()">
 
     <div *ngIf="title"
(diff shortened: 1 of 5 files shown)


projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.ts
```diff
@@ -68,6 +68,7 @@ export class CollapsibleSectionComponent
   @Input() expanded = false;
   @Input() disabled = false;
   @Input() divided = true;
+  @Input() showLeftSideBorder = false;
```
galloween, line 71: lets not make it a separate input, but extend the [options] input instead (and CollapsibleOptions interface)

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
```diff
@@ -39,7 +40,8 @@
        [attr.id]="panelID"
        [attr.aria-labelledby]="panelID+'-header'"
        [attr.data-collapsible]="!!collapsible"
-       [attr.data-expanded]="!collapsible || startsExpanded ? null : !!expanded">
+       [attr.data-expanded]="!collapsible || startsExpanded ? null : !!expanded"
+       [class.show-border]="showLeftSideBorder">
```
galloween, line 44: lets keep to convention (specific to this component) - and use attributes instead of class

galloween (changes requested):
looks good, just 2 little adjustments... :)

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
galloween: also, its enough to have 1 attribute - on the <section> element

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.ts
galloween: actually, thinking about it, we dont need the showLeftSideBorder at all - we can use the indicatorColor. its not used for anything else, so we can say - if indicatorColor is present, show the border

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
```diff
@@ -50,6 +50,16 @@ $bcp-border-color: $border-color-light;
...
+
+  &[show-border='true'] {
+    .bcp-header {
+      border-left: 16px solid var(--bcp-color);
+    }
+
+    .bcp-panel {
+      border-left: 16px solid rgba(var(--bcp-color-rgb), 0.3);
+    }
+  }
```
galloween, line 62: lets move this to the bottom of the file - there is  "Options" section

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.html
```diff
@@ -2,7 +2,8 @@
          [attr.aria-disabled]="disabled || null"
          [attr.data-collapsible]="!!collapsible"
          [attr.data-expanded]="!collapsible ? null : !!expanded"
-         [attr.data-divided]="!!divided">
+         [attr.data-divided]="!!divided"
+         [attr.show-border]="!!options.indicatorColor">
```
galloween, line 6: custom attributes by html5 specification should have 'data-' prefix, we also follow this convention

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
MaxDsr: @galloween , should I also add it with `::ng-deep` like it is there ? 

`.bcp-section[data-divided='false']::ng-deep`

There aren't angular components inside this component so I don't think if it is needed.

galloween: you actually need to use this structure:

```
.bcp-section[data-show-border='true']::ng-deep {
  & > .bcp-header {
    border: ....
  }

  & > .bcp-panel {
    borde: ....
  }
}
```


the reason is, sometimes we have CollapsibleSection **_inside_** CollapsibleSection, so with `>` we make sure only this section is affected. `::ng-deep` here is not really needed, it just helps to simplify the compiled css selectors...

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.stories.ts
```diff
@@ -94,15 +91,14 @@ const note = `
...
   [title] | string | section title | &nbsp;
   [description] | string | section description (subtitle) | &nbsp;
   [options] | CollapsibleOptions | additional options, among which: <br> **options.headerTranscludeStopPropagation** \
   - set to true to prevent click event propagation from content transcluded in header (for example, to prevent buttons\
      opening/closing the panel)<br>**options.indicatorColor** - will add \`--bcp-color\` and \`--bcp-color-rgb\` css \
      variables, that you can use for custom color mods in your feature css, \
      for example: \`color: var(--bcp-color)\`, or for rgba color with opacity \
-     - \`color: rgba(var(--bcp-color-rgb), 0.2)\` \
+     - \`color: rgba(var(--bcp-color-rgb), 0.2)\` \. Also will add border on the left of the section with this color.
      | COLLAPSIBLE<sub>-</sub>OPTIONS<sub>-</sub>DEF
```
galloween, line 102: lets just leave "Will add border... blah blah"
the explanation about the variables is not relevant anymore (they were created for the borders which are now built-in)

galloween (approved):

PR #1734, opened 2020-09-22
maxdsr/single select add no data message and keep height

MaxDsr:
- add _**no results found**_ message when nothing from select search was found.
- keep the current height of the list when user types in search box and results list is smaller


galloween (changes requested):
@MaxDsr 

the height calculation needs to be smarter. it should be no more than the maxHeight, otherwise the list gets too high (Math.min from this.maxHeight and getListHeight). also its not enough to have this calculation on init (i suggest you put it in the end of updateLists method). and you need listOptions.length, not options (listOptions is the viewModel, options is the data model)

+ you need to hide the 'none' option

as for styles, if you add classes "option-select option", it will have most of what you need. you will just need to modify the text color to $grey-600;

also for the text, it needs to come from translation.
the translations are in hibob-web-client:  projects/shared/src/assets/i18n/locale-en.json
you can add to bob-style > lists > empty_state (for example).
then it can be used in bob-style (with the translate pipe or .instant method). The translation will not be available immediately, but its ok.

image

image

galloween (commented):
one last little thing - lets rename the var and method to reflect its relation to min-height
+ maybe add changes for multi-list to same PR ?

galloween (approved):

PR #1738, opened 2020-09-24
Maxdsr/remove shadow and add dark overlay on confirmation dialog

MaxDsr:
remove box-shadow from the top of the mat-dialog-actions element and replace it with border

add overlay background for whole dialog when confirmation is poped-up

### DIFF
--- a/projects/ui-framework/src/lib/style/dialog.scss
+++ b/projects/ui-framework/src/lib/style/dialog.scss
@@ -102,10 +102,6 @@ html.cdk-global-scrollblock {
           background-color: black(0.3);
           z-index: 1;
         }
-
-        & + .dialog-footer .mat-dialog-actions {
-          border-top: 1px solid $grey-500;
-        }
       }
 
       .dialog-content {
@@ -157,6 +153,19 @@ html.cdk-global-scrollblock {
       background-color: $white;
       box-shadow: 0 -3px 10px rgba(0, 0, 0, 0.2);
       transform: translateY(-100%);
+
+      & + .mat-dialog-actions {
+        box-shadow: 0 -1px 0 0 $border-color;
+
+        [mat-dialog-footer-left] {
+          @include disabled-state {
+            *,
+            :before {
+              pointer-events: none;
+            }
+          }
+        }
+      }
     }
 
     .mat-dialog-actions {
--- a/projects/ui-framework/src/lib/style/dialog.scss
+++ b/projects/ui-framework/src/lib/style/dialog.scss
@@ -50,13 +50,6 @@ html.cdk-global-scrollblock {
       padding: times8(5) times8(5) times8(2);
       user-select: none;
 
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/style/dialog.scss
```diff
@@ -91,12 +99,12 @@ html.cdk-global-scrollblock {
...
           @include position(0 0 0 0);
           pointer-events: all;
-          background-color: white(0.4);
+          background-color: black(0.3);
           z-index: 1;
         }
 
         & + .dialog-footer .mat-dialog-actions {
-          box-shadow: 0 -3px 5px rgba(0, 0, 0, 0.1);
+          border-top: 1px solid $grey-500;
```
galloween, line 107: for border-color we have a special variable: $border-color

projects/ui-framework/src/lib/style/dialog.scss
```diff
@@ -46,9 +46,17 @@ html.cdk-global-scrollblock {
...
       padding: times8(5) times8(5) times8(2);
       user-select: none;
 
+      &.show-confirmation:before {
+        @include pseudo;
+        @include position(0 0 0 0);
+        background-color: black(0.3);
+        z-index: 1;
+      }
+
```
galloween, line 59: i think this can be combined with the other shadow into 1 ruleset, in the end of the file:

```
.b-dialog-panel {
  .show-confirmation {
    user-select: none;
    pointer-events: none;
    position: relative;

    &:before {
      @include pseudo;
      @include position(0 0 0 0);
      pointer-events: all;
      background-color: black(0.3);
      z-index: 1;
    }
  }
}
```

galloween (commented):
good! see suggestion for a little optimization

galloween (commented):
looks good, see suggestion for a little optimisation

projects/ui-framework/src/lib/style/dialog.scss
galloween: actually, i think its better if all confirmation-related rules are grouped in one block...

also, lets move the .mat-dialog-actions border/shadow rule to .confirmation-message + add disabled on the [mat-dialog-footer-left] ng-content:

around line 146:

```
.confirmation-message {
     ....

      & + .mat-dialog-actions {
        border-top: 1px solid $grey-500; // change to shaddow

        [mat-dialog-footer-left] {
          @include disabled-state {
            *,
            :before {
              pointer-events: none;
            }
          }
        }
      }
    }
```

projects/ui-framework/src/lib/style/dialog.scss
galloween: but! if you look closely, when the border appears, it moves the dialog a little bit, 
so i think its better to stay with the shadow:

`box-shadow: 0 -1px 0 0 $border-color;`

galloween (approved):

PR #1764, opened 2020-10-06
fix global link reset class

doroncy:
b-reset-link was missing hover class


projects/ui-framework/src/lib/style/atomic.scss
```diff
@@ -299,6 +299,9 @@ $acss-dirs: top right bottom left;
   font: inherit;
   color: inherit;
   text-decoration: inherit;
+  &:hover {
+    text-decoration: inherit;
+  }
```
galloween, line 304: oh yes :)

galloween (approved):

PR #1775, opened 2020-10-12
Pavel/misc - fonts, rte, chain-select, multi-list-and-chips, selects

galloween:


projects/ui-framework/src/lib/services/utils/functional-utils.ts
```diff
@@ -141,6 +156,24 @@ export const getType = (smth: any): string =>
     ? 'DOM node'
     : String(typeof smth);
 
+export const getEmptyOfSameType = <T = unknown>(item: T): T => {
```
danidin, line 159: I think we need `string` and `number` to return `null`, not `undefined`

projects/ui-framework/src/lib/lists/chain-select/chain-select.component.ts
```diff
@@ -25,19 +46,36 @@ export class ChainSelectComponent implements OnInit {
...
-      this.chainLinkList = [{}];
+      if (!this.chainLinkList.length) {
+        this.addChainLink();
+      }
+    }
+  }
+
+  ngOnInit(): void {
+    if (!this.chainLinkList) {
+      this.emptyItem = undefined;
```
danidin, line 71: I think the default should be `null`, because in out current typing system that will pass as any typed item whereas `undefined` will not

projects/ui-framework/src/lib/services/utils/functional-utils.ts
galloween: well, from semantic point of view, undefined is more appropriate - as in 'value not defined', 
vs null's 'intentional absence of any value' - which is a value.

so i'd keep it undefined here, but `||` it to null in the component.

projects/ui-framework/src/lib/lists/chain-select/chain-select.component.ts
galloween: bob-style has no problem with  `private emptyItem: any = undefined;`

and as for bob2, you can still pass `null` to [selectedItemList] (and if its the first element in the array it will make emptyItem to equal null as well...

ok, i will change the default to null, maybe in this case its semantically appropriate (for example when used to reset select's value to empty - it may actually ignore 'undefined' in value input)

PR #1792, opened 2020-10-22
smaller quick filter bar loader, vertical separator

harellevy:


projects/ui-framework/src/lib/search/quick-filter/quick-filter-bar.component.scss
```diff
@@ -182,4 +190,18 @@
...
+  .smaller {
+    .bar-filters:empty {
+      &:before {
+        height: $wrap-height-smaller;
+      }
+    }
+    .bfe-wrap {
+      &:before,
+      &:after {
+        @include quick-filter-separator-pseudo-smaller;
```
galloween, line 203: you just need to adjust the bottom here, no need to duplicate mixin code

projects/ui-framework/src/lib/search/quick-filter/quick-filter-bar.component.html
```diff
@@ -1,42 +1,44 @@
...
-        *ngIf="showResetFilter"
-        class="reset-filters"
-        role="button"
-        [icon]="icons.refresh"
-        [color]="iconColor.normal"
-        [size]="iconSize.large"
-        [hasHoverState]="true"
-        (click)="onReset()">
-</b-icon>
+<div [class.smaller]="size === 'smaller'">
```
galloween, line 1: it would be better if you used: 

`@HostBinding('attr.data-size') @Input() size = FormElementSize.regular;`

and then all the modifications in:

`:host[data-size="smaller"] { ... .bar-suffix { ... } ...  }`

(no need for extra wrappers)

galloween (changes requested):
i try to use data-attributes on host elements for things like type, size and other visual modifiers in other components, so it's better if this one uses the same approach....

galloween (approved):

PR #1814, opened 2020-11-03
add description tooltips of headers for lists

MaxDsr:
Add tooltip with description for group headers in the single list.
Add description field for `SelectGroupOption` interface.
Add distinct hover state class that doesn't change tooltip icon hover state with its hover state.
Add `z-index` for tooltip icons. It doesn't triggers because of header block's `z-index`.
Add tests.

All changes related to task: https://app.asana.com/0/1159939249943132/1198969695880245/f

### DIFF
--- a/projects/ui-framework/src/lib/lists/multi-list/multi-list.component.html
+++ b/projects/ui-framework/src/lib/lists/multi-list/multi-list.component.html
@@ -31,7 +31,7 @@
             'selected': header.selected,
             'indeterminate': header.indeterminate
           }"
-           class="option-select header has-hover">
+           class="option-select header has-distinct-hover">
         <span (click)="headerClick(header)"
               class="header-collapse-trigger">
         </span>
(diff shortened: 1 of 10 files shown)


projects/ui-framework/src/lib/lists/multi-list/multi-list.component.html
```diff
@@ -31,7 +31,7 @@
             'selected': header.selected,
             'indeterminate': header.indeterminate
           }"
-           class="option-select header has-hover">
+           class="option-select header has-distinct-hover">
```
galloween, line 34: lets leave the has-hover class, it will not interfere with the has-hover on the description icon, but its just a common class thats used a lot.

projects/ui-framework/src/lib/lists/single-list/single-list.component.scss
```diff
@@ -153,6 +153,7 @@ $list-max-items: 8;
     margin-right: -10px;
     margin-left: auto;
     visibility: hidden;
+    z-index: 2;
```
galloween, line 156: also add cursor: default here

galloween (approved):
<s>all good, i think its better to leave the has-hover class, because its just a convention and it will not interfere with the other has-hover on the icon</s>

never mind the above comment, i see it does interfere, so all good :) thanks

projects/ui-framework/src/lib/lists/multi-list/multi-list.component.html
galloween: oh i see it does interfere, ok, leave with the distinct!

PR #1819, opened 2020-11-08
angular 10.2

harellevy:


tsconfig.base.json
```diff
@@ -6,7 +6,7 @@
     "outDir": "./dist/out-tsc",
     "sourceMap": true,
     "declaration": false,
-    "module": "esnext",
+    "module": "es2020",
```
doroncy, line 9: if it wont work maybe do not do this change here

doroncy (approved):

PR #1827, opened 2020-11-12
David/multi list and list

SkyWallkeRed:
IN DEVELOPMENT

added a new component to Lists b-multi-list-and-list

https://www.figma.com/file/e5kH0ehswD91XqrcvwdIy0/Flows?node-id=0%3A1

### DIFF
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
@@ -4,7 +4,6 @@
                 [mode]="mode"
                 [listActions]="listActions"
                 [options]="listOptions"
-                [optionsDefault]="optionsDefault"
                 [maxHeight]="listElHeight*6"
                 [showSingleGroupHeader]="showSingleGroupHeader"
                 [startWithGroupsCollapsed]="startWithGroupsCollapsed"
@@ -16,7 +15,8 @@
 
 <div class="mlac-list">
 
-  <b-basic-list [type]="type" [items]="selectedListOptions"
+  <b-basic-list [type]="type"
+                [items]="selectedListOptions"
                 [config]="{
                   showActionOnHover: showActionOnHover,
                   maxLines: maxLines
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
@@ -1,4 +1,8 @@
 <div class="mlac-list">
+  <p class="mlac-list-label"
+     [attr.id]="listID"
+     [attr.aria-hidden]="!listLabel"
+     *ngIf="listLabel">{{listLabel}}</p>
 
   <b-multi-list [attr.aria-labelledby]="listLabel ? listID : null"
                 [mode]="mode"
@@ -14,6 +18,10 @@
 </div>
 
 <div class="mlac-list">
+  <p class="mlac-list-label"
+     [attr.id]="listID"
+     [attr.aria-hidden]="!listLabel"
+     *ngIf="selectedLabel">{{selectedLabel}}</p>
 
   <b-basic-list [type]="type"
                 [items]="selectedListOptions"
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
@@ -1,5 +1,5 @@
-<div class="mlal-list">
-  <p class="mlal-list-label"
+<div class="mlac-list">
+  <p class="mlac-list-label"
      [attr.id]="listID"
      [attr.aria-hidden]="!listLabel"
      *ngIf="listLabel">{{listLabel}}</p>
@@ -17,8 +17,8 @@
   </b-multi-list>
 </div>
 
-<div class="mlal-list">
-  <p class="mlal-list-label"
+<div class="mlac-list">
+  <p class="mlac-list-label"
      [attr.id]="listID"
      [attr.aria-hidden]="!listLabel"
      *ngIf="selectedLabel">{{selectedLabel}}</p>
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
@@ -1,5 +1,5 @@
-<div class="mlac-list">
-  <p class="mlac-list-label"
+<div class="mlal-list">
+  <p class="mlal-list-label"
      [attr.id]="listID"
      [attr.aria-hidden]="!listLabel"
      *ngIf="listLabel">{{listLabel}}</p>
@@ -17,8 +17,8 @@
   </b-multi-list>
 </div>
 
-<div class="mlac-list">
-  <p class="mlac-list-label"
+<div class="mlal-list">
+  <p class="mlal-list-label"
      [attr.id]="listID"
      [attr.aria-hidden]="!listLabel"
      *ngIf="selectedLabel">{{selectedLabel}}</p>
@@ -33,19 +33,17 @@
 
     <ng-container *bBasicListAction="let item=item">
 
-      <b-menu *ngIf="listViewConfig.rowAction.actionType === 'menu'" [menu]="item.menu">
+      <b-menu *ngIf="listViewConfig.rowAction.menu" [menu]="item.menu">
         <b-square-button menu-trigger
                          [type]="buttonType"
-                         [icon]="item.actionIcon"
-                         [color]="iconColor">
+                         [icon]="item.actionIcon">
         </b-square-button>
       </b-menu>
 
-      <b-square-button *ngIf="listViewConfig.rowAction.actionType === 'icon'"
+      <b-square-button *ngIf="listViewConfig.rowAction.icon && !listViewConfig.rowAction.menu"
                        [disabled]="item.disabled"
                        [type]="buttonType"
                        [icon]="item.actionIcon"
-                       [color]="iconColor"
                        (clicked)="onListRowRemove(item)"
       >
       </b-square-button>
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
@@ -1,0 +1,48 @@
+<div class="mlac-list">
+
+  <b-multi-list [attr.aria-labelledby]="listLabel ? listID : null"
+                [mode]="mode"
+                [listActions]="listActions"
+                [options]="listOptions"
+                [optionsDefault]="optionsDefault"
+                [maxHeight]="listElHeight*6"
+                [showSingleGroupHeader]="showSingleGroupHeader"
+                [startWithGroupsCollapsed]="startWithGroupsCollapsed"
+                [min]="min"
+                [max]="max"
+                (selectChange)="onListChange($event)">
+  </b-multi-list>
+</div>
+
+<div class="mlac-list">
+
+  <b-basic-list [type]="type" [items]="selectedListOptions"
+                [config]="{
+                  showActionOnHover: showActionOnHover,
+                  maxLines: maxLines
+                }"
+                [emptyStateConfig]="emptyStateConfig">
+
+    <ng-container *bBasicListAction="let item=item">
+
+      <b-menu *ngIf="listViewConfig.rowAction.actionType === 'menu'" [menu]="item.menu">
+        <b-square-button menu-trigger
+                         [type]="buttonType"
+                         [icon]="item.actionIcon"
+                         [color]="iconColor">
+        </b-square-button>
+      </b-menu>
+
+      <b-square-button *ngIf="listViewConfig.rowAction.actionType === 'icon'"
+                       [disabled]="item.disabled"
+                       [type]="buttonType"
+                       [icon]="item.actionIcon"
+                       [color]="iconColor"
+                       (clicked)="onListRowRemove(item)"
+      >
+      </b-square-button>
+
+    </ng-container>
+  </b-basic-list>
+
+</div>
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
@@ -46,7 +46,8 @@ b-empty-state {
 }
 
 :host ::ng-deep table {
-  min-height: 360px;
+  min-height: 300px;
+  overflow-y: scroll;
 
   tbody {
     width: 100%;
@@ -56,11 +57,12 @@ b-empty-state {
       width: 100%;
     }
 
-    //td:nth-child(2) {
-    //  alignment: right;
-    //  position: absolute;
-    //  right: 100px;
-    //  padding-top: 14px;
-    //}
+    td:nth-last-child(2) {
+      text-align: right;
+    }
+
+    td:last-child {
+      width: 10px;
+    }
   }
 }
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
@@ -9,13 +9,13 @@ $gap: 20px;
   --translation-all: '(all)';
 }
 
-.mlac-list {
+.mlal-list {
   display: flex;
   flex-direction: column;
   width: calc(50% - #{$gap / 2});
 }
 
-.mlac-list-label {
+.mlal-list-label {
   @include bFormElement-label(false);
 }
 
@@ -24,8 +24,9 @@ b-basic-list {
   min-height: 300px;
   max-height: 360px;
   background-color: white;
-  overflow-y: scroll;
-  box-shadow: none;
+  overflow-y: auto;
+  border: solid 1px $border-color;
+  border-radius: $border-radius;
 }
 
 b-multi-list {
@@ -45,24 +46,14 @@ b-empty-state {
   padding: inherit;
 }
 
-:host ::ng-deep table {
-  min-height: 300px;
-  overflow-y: scroll;
+:host ::ng-deep b-basic-list > table {
 
   tbody {
-    width: 100%;
+    box-shadow: none !important;
 
-    tr {
-      display: table;
-      width: 100%;
-    }
-
-    td:nth-last-child(2) {
+    .bbl-list-label + .bbl-list-label {
       text-align: right;
     }
-
-    td:last-child {
-      width: 10px;
-    }
   }
+
 }
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
@@ -19,14 +19,13 @@ $gap: 20px;
   @include bFormElement-label(false);
 }
 
-
-b-basic-list{
+b-basic-list {
   flex-grow: 5;
   min-height: 300px;
   max-height: 360px;
-  box-shadow: none;
   background-color: white;
   overflow-y: scroll;
+  box-shadow: none;
 }
 
 b-multi-list {
@@ -37,11 +36,6 @@ b-multi-list {
   background-color: white;
   overflow-y: scroll;
 }
-//
-//:host ::ng-deep b-basic-list {
-//  min-height: 300px;
-//  max-height: 360px;
-//}
 
 b-empty-state {
   align-self: center;
@@ -49,4 +43,24 @@ b-empty-state {
   justify-content: center;
   text-align: center;
   padding: inherit;
+}
+
+:host ::ng-deep table {
+  min-height: 360px;
+
+  tbody {
+    width: 100%;
+
+    tr {
+      display: table;
+      width: 100%;
+    }
+
+    //td:nth-child(2) {
+    //  alignment: right;
+    //  position: absolute;
+    //  right: 100px;
+    //  padding-top: 14px;
+    //}
+  }
 }
--- a/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
+++ b/projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
@@ -1,0 +1,52 @@
+@import '../../style/common-imports';
+
+$gap: 20px;
+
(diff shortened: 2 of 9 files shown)


projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+      if (op.selected) {
+        const option: ListRow = {
+          id: op.id,
+          label: [op.value],
+          disabled: op.disabled,
+        };
+        // set row strong tag if any
+        if (op.strongParent) {
+          option.label = [`<strong>${optionGroup.groupName}</strong> - ${op.value}`];
+        }
```
galloween, line 164: i think this should be default hardcoded behaviour, without strongParent toggle.

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+        // set sub value in any
+        if (op.subValue) {
+          option.label = [...option.label, op.subValue];
+          // set empty ele to support table view
+        } else if (this.listViewConfig.rowTableView) {
+          option.label = [...option.label, ''];
```
galloween, line 170: needs to be empty space if you really mean to put an empty column, otherwise its falsy and wont achieve anything

'' => ' '

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+        }
+        // set row start icon if any
+        if (this.listViewConfig?.rowStartIcon) {
+          option.icon = this.listViewConfig.rowStartIcon;
+        }
+        // set row action icon if any
+        if (this.listViewConfig?.rowAction) {
+          option.actionIcon = this.listViewConfig?.rowAction?.icon;
+        }
```
galloween, line 179: you dont really need the if's. you can just assign, and its undefined - its same thing as not assigning.

so it will be a bit more compact if you just use map instead of the forEach, and just map to an object: 

```
{
  id: op.id,
  label: [`<strong>${optionGroup.groupName}</strong> - ${op.value}`, op.subValue || ' '],
  icon: this.listViewConfig.rowStartIcon,
  actionIcon = this.listViewConfig?.rowAction?.icon,
 ... etc ...
}
```

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+export interface ListViewConfig {
+  rowStartIcon?: Icons;
+  rowAction?: {
+    actionType: RowActionType;
+    icon: Icons;
+    menu?: MenuItem[];
+  };
+  rowTableView?: boolean;
+}
+
```
galloween, line 46: i think you dont need actionType - you either have menu or not - its enough to decide what template to use

galloween: also not sure what rowTableView is for - you are supposed to have same amount of labels for every item.
if you think there may be cases, where some items have subValue and some not - just add the empty cell for those items that dont have it (below)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+export interface ListRow extends BasicListItem {
+  icon?: Icons;
```
galloween, line 48: icon already exists in BasicListItem, so no need to repeat

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+  actionIcon?: Icons;
+  menu?: MenuItem[];
+  id: number | string;
+  disabled?: boolean;
+}
+
+@Component({
+  selector: 'b-multi-list-and-list',
+  templateUrl: './multi-list-and-list.component.html',
+  styleUrls: ['./multi-list-and-list.component.scss'],
```
galloween, line 58: if you are re-using classes and styles from the MLAC component, just add an import to its scss file here, before the multi-list-and-list

for classes, unique to this component, lets use MLAL prefix (the C in the MLAC was chips)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    private translate: TranslateService,
+    private DOM: DOMhelpers,
+  ) {
+    this.listActions = { ...MULTI_LIST_LIST_ACTIONS_DEF };
+  }
+
+  readonly buttonType: ButtonType = ButtonType.tertiary;
+
+  @Input() icon: Icons = Icons.three_dots;
+  @Input() iconColor: IconColor;
```
galloween, line 74: where do these go?
iconColor i saw in template, but icon is not used i think.

also, lets have iconColor hardcoded for now. there is no need to have to pass it all the time

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
SkyWallkeRed: the design have 3 states one is with strong one is without

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
galloween: the `|| ' '` trick is only needed if there might be some items with subValue and some without. 
If all of them are expected to be the same, you can do this:
```
[`<strong>${optionGroup.groupName}</strong> - ${op.value}`, ...(op.subValue ? [op.subValue] : [])]
```
or
```
[`<strong>${optionGroup.groupName}</strong> - ${op.value}`, op.subValue].filter(Boolean)
```

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.module.ts
```diff
@@ -0,0 +1,26 @@
...
+    CommonModule,
+    MultiListModule,
+    EmptyStateModule,
+    TranslateModule,
+    BasicListModule,
+    ButtonsModule,
+    MenuModule,
+    IconsModule,
+    // BasicListModule,
+    // ButtonsModule,
```
galloween, line 21: remove

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.module.ts
```diff
@@ -0,0 +1,26 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { BasicListModule, ButtonsModule, IconsModule, MenuModule, MultiListModule } from 'bob-style';
+import { EmptyStateModule } from 'bob-style';
+import { TranslateModule } from '@ngx-translate/core';
+import { MultiListAndListComponent } from './multi-list-and-list.component';
+// import { BasicListModule, ButtonsModule } from 'bob-style';
```
galloween, line 7: remove

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.module.ts
```diff
@@ -0,0 +1,26 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { BasicListModule, ButtonsModule, IconsModule, MenuModule, MultiListModule } from 'bob-style';
+import { EmptyStateModule } from 'bob-style';
```
galloween, line 4: imports from within same library should always be relative

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
```diff
@@ -0,0 +1,68 @@
...
+
+:host ::ng-deep table {
+  min-height: 300px;
+  overflow-y: scroll;
+
+  tbody {
+    width: 100%;
+
+    tr {
+      display: table;
```
galloween, line 56: you cant do this. it breaks the table

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
```diff
@@ -0,0 +1,68 @@
...
+      width: 100%;
+    }
+
+    td:nth-last-child(2) {
+      text-align: right;
```
galloween, line 61: this is unpredictable.
if you mean to address second label cell, use:
```.bbl-list-label + .bbl-list-label```

File: projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss [hunk 13]
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    td:last-child {
+      width: 10px;
```
galloween, line 65: again, this is unpredictable.
if you mean to address the cell with the action, use `.bbl-list-action-container`
but why would you need to set its width to 10px????

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
```diff
@@ -0,0 +1,68 @@
...
+  align-self: center;
+  margin: auto;
+  justify-content: center;
+  text-align: center;
+  padding: inherit;
+}
+
+:host ::ng-deep table {
+  min-height: 300px;
+  overflow-y: scroll;
```
galloween, line 50: bad idea. 
1) remove overflow from table, you already have it on the container - b-basic-list
2) remove min-height from table; to manage the border, remove box-shadow from `.bbl-lis body` and add border and border-radius to `b-basic-list`

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+  @Input() selectedLabel: string;
+  @Input() showSingleGroupHeader = false;
+  @Input() startWithGroupsCollapsed = true;
+  @Input() listActions: ListFooterActions;
+  // use this group
+  // @Input() mode: SelectMode = SelectMode.checkGroups;
+  // for testing
+  @Input() mode: SelectMode = SelectMode.classic;
+
+  @Input() public type: BasicListType = BasicListType.primary;
```
galloween, line 89: this should be either hardcoded or omitted. no need to config it at the moment

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
```diff
@@ -0,0 +1,68 @@
...
+  box-shadow: none;
+}
+
+b-multi-list {
+  flex-grow: 5;
+  min-height: 300px;
+  max-height: 360px;
+  box-shadow: none;
+  background-color: white;
+  overflow-y: scroll;
```
galloween, line 37: remove overflow. it can manage its own overflow

galloween: (and better just keep the styles from MLAC, no need to invent again)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
```diff
@@ -0,0 +1,68 @@
...
+.mlac-list-label {
+  @include bFormElement-label(false);
+}
+
+b-basic-list {
+  flex-grow: 5;
+  min-height: 300px;
+  max-height: 360px;
+  background-color: white;
+  overflow-y: scroll;
```
galloween, line 27: use `auto` instead of `scroll` so that scrollbar does not show when not needed

galloween (changes requested):
see comments

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
galloween: can you add link to design in PR description so that i can take a look?

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
galloween: also it will have no affect on the table, it will still grow to whatever size it wants

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,222 @@
...
+    this.options = listChange.getSelectGroupOptions();
+    this.optionsToList(this.options);
+    this.emitChange();
+  }
+
+  private optionsToList(options: SelectGroupOption[]): ListRow[] {
+    const listItems: ListRow[] = [];
+
+    options.forEach(optionGroup => optionGroup.options.forEach(op => {
+
```
galloween, line 154: when all the group is selected, are you sure you want to add all options separately?
(just checking)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
SkyWallkeRed: https://www.figma.com/file/e5kH0ehswD91XqrcvwdIy0/Flows?node-id=0%3A1

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
SkyWallkeRed: it will be inited with mode: SelectMode = SelectMode.checkGroups; so there is no option to select more then one in group, i dod that to support a generic structure..

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
SkyWallkeRed: i did that to support more generic use cases, its not needed in the design, (good to have ?)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
```diff
@@ -0,0 +1,54 @@
+<div class="mlal-list">
+  <p class="mlal-list-label"
+     [attr.id]="listID"
+     [attr.aria-hidden]="!listLabel"
+     *ngIf="listLabel">{{listLabel}}</p>
+
+  <b-multi-list [attr.aria-labelledby]="listLabel ? listID : null"
+                [mode]="mode"
```
galloween, line 8: when i meant "re-use/keep styles and classes from MLAchips" i meant :
- keep the mlac classes as much as possible, 
- dont duplicate **none** of the scss, instead just use the same file in component metadata:
`styleUrls: ['..path-to/mlac/mlac.style.file.scss',  './multi-list-and-list.component.scss'],`
- for unique stuff (that doesnt have a similar thing in MLAchips), use the new mlal prefix

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.scss
```diff
@@ -0,0 +1,59 @@
+@import '../../style/common-imports';
+
```
galloween, line 2: alternatively, you could also import the MLAC file here

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+  public onListChange(listChange: ListChange): void {
+    this.options = listChange.getSelectGroupOptions();
+    this.optionsToList(this.options);
+    this.emitChange();
+  }
+
+  private optionsToList(options: SelectGroupOption[]): ListRow[] {
+    const listItems: ListRow[] = [];
+    options.forEach(optionGroup => optionGroup.options.map(op => {
+      if (op.selected) {
```
galloween, line 141: indeed in this case map doesnt make any sense, as you are not mapping but pushing to external array (my advise doesnt fit here).
so stay with forEach :)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+      this.cd.detectChanges();
+    }
+
+    if (!this.cd['destroyed']) {
+      this.cd.detectChanges();
+    }
+  }
+
+  public onListChange(listChange: ListChange): void {
+    this.options = listChange.getSelectGroupOptions();
```
galloween, line 133: with the better approach than mlac, with separating options from value,
you dont need to save the options, they are (basically) the same as you got as input,

what you need is:

this.value = listChange.delectedIDs;
this.optionsToList( listChange.getSelect<u>**ed**</u>GroupOptions(), this.value  );

selectED group options is - you guessed it - just the selected options,
second argument (value) is explained below

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+  @Input() public showActionOnHover = false;
+  @Input() public maxLines: number = null;
+
+  @Input() public emptyStateConfig: EmptyStateConfig = { icon: Icons.three_dots };
+  @Input() public listViewConfig: ListViewConfig;
+
+  @Input() min: number;
+  @Input() max: number;
+
+  @Input() options: SelectGroupOption[] = [];
```
galloween, line 86: i think we can do something better than the MLAC.

lets add `@Input() value: (string|number)[];`  <- list of selected IDs, same as multi-list component's value

this way we can separate options from value, it will be much nicer

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+  }
+
+  public onListChange(listChange: ListChange): void {
+    this.options = listChange.getSelectGroupOptions();
+    this.optionsToList(this.options);
+    this.emitChange();
+  }
+
+  private optionsToList(options: SelectGroupOption[]): ListRow[] {
```
galloween, line 138: in the new approach, you need both options AND value:

 private optionsToList( options: SelectGroupOption[], value: (string|number)[] ) {}

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
galloween: and the selected check becomes:

if ( isArray(this.value) ?  value.includes(op.id) : op.selected  ) { ... }

isArray check is for the cases, where options is the only source of value. but after the first change (see below), we will have value and from then on, we can ignore the options

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+  @Input() max: number;
+
+  @Input() options: SelectGroupOption[] = [];
+  @Input() optionsDefault: SelectGroupOption[];
+  @Output() selectChange: EventEmitter<ListChange> = new EventEmitter<ListChange>();
+  @Output() menuAction: EventEmitter<any> = new EventEmitter<any>();
+  @Output() clicked: EventEmitter<MouseEvent> = new EventEmitter<MouseEvent>();
+
+  public type: BasicListType = BasicListType.primary;
+  public listOptions: SelectGroupOption[] = [];
```
galloween, line 93: you will not need this one (see below)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
```diff
@@ -0,0 +1,54 @@
...
+                [listActions]="listActions"
+                [options]="listOptions"
+                [maxHeight]="listElHeight*6"
+                [showSingleGroupHeader]="showSingleGroupHeader"
+                [startWithGroupsCollapsed]="startWithGroupsCollapsed"
+                [min]="min"
+                [max]="max"
+                (selectChange)="onListChange($event)">
```
galloween, line 16: you will also need to bind [value]="value" (see other comments for explanation)

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+        opt.selected = false;
+      }
+    });
+    this.selectedListOptions = this.selectedListOptions.filter((listOp: any) => listOp.id !== item.id);
+    return (this.options = options);
+  }
+
+  private emitChange(): void {
+    this.selectChange.emit(new ListChange(this.listOptions));
+  }
```
galloween, line 180: ListChange can take second argument - the selected IDs, so do:

`new ListChange(this.options, this.value)`

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.ts
```diff
@@ -0,0 +1,186 @@
...
+    const options: SelectGroupOption[] = cloneArray(this.options);
+    options.find((group: SelectGroupOption) => {
+      const opt = group.options.find((o: SelectOption) => o.id === item.id);
+      if (opt) {
+        opt.selected = false;
+      }
+    });
+    this.selectedListOptions = this.selectedListOptions.filter((listOp: any) => listOp.id !== item.id);
+    return (this.options = options);
+  }
```
galloween, line 176: so in this case, with the new approach, we dont need to touch the options, but instead:
- remove item.id from value
- call this.optionsToList(this.options, this.value)
- detectChanges

galloween (commented):
ok there is something we can (possibly) do better (maybe) 
if you chose to go your own way with this - its also ok

projects/ui-framework/src/lib/lists/multi-list-and-list/multi-list-and-list.component.html
SkyWallkeRed: b-basic-list dont support [value] as input

galloween (approved):

oridanus:
party

PR #1830, opened 2020-11-16
Daniel/delete confirmation dialog

malkind:


projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.html
```diff
@@ -0,0 +1,15 @@
...
+  <div b-dialog-content>
+    <p *ngIf="config.message" [innerHTML]="config.message"></p>
+
+    <b-input #bInput
+      [label]="config.confirmationData.label"
+      [placeholder]="config.confirmationData.placeholder"
+      [required]="true"
+      [hintMessage]="config.confirmationData.hintMessage"
+      [errorMessage]="errMessage"
+      (inputEvents)="onInputEvent($event)">
```
galloween, line 12: it would be nice if there were fallbacks for everything - label, placeholder, hint

projects/ui-framework/src/lib/popups/confirmation-dialog/confirmation-dialog.interface.ts
```diff
@@ -5,9 +5,18 @@ export interface ConfirmationDialogConfig {
...
   ok: DialogButton;
   cancel?: DialogButton;
 }
+
+export interface ConfirmationData {
+  confirmationText: string;
+  label?: string;
+  placeholder?: string;
+  hintMessage?: string;
+  errorMessage?: string;
```
galloween, line 21: if  label, placeholder, hint are optional - you need to provide default fallbacks in the component
maybe confirmationText should also have a fallback and be optional

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  readonly icons = Icons;
+
+  errMessage = '';
+
+  ngOnInit() {
+    const okCb = this.config.buttonConfig.ok.action;
+    this.config.buttonConfig.ok.action = () => {
+      if (this.isValid()) {
+        return okCb();
+      }
```
galloween, line 29: why do you need to save it to const? 
you can just `return  this.config.buttonConfig.ok.action()`

galloween: maybe its also a good idea to check    if  `isFunction(this.config.buttonConfig.ok.action)`

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    public dialogRef: MatDialogRef<DeleteConfirmationDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public config: ConfirmationDialogConfig
+  ) {}
+
+  @ViewChild('bInput') bInput: InputComponent;
+
+  readonly buttonSize = ButtonSize;
+  readonly buttonType = ButtonType;
+  readonly icons = Icons;
+
```
galloween, line 21: i dont think you use these

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+  }
+
+  ngOnDestroy(): void {
+    this.dialogRef.close();
+  }
+
+  isValid(): boolean {
+    return this.bInput.value.length > 0 && this.errMessage.length === 0;
+  }
```
galloween, line 42: you should have only 1 validation method 
`this.bInput.value. === this.config.confirmationData.confirmationText`

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  onInputEvent(input: InputEvent) {
+    if (this.config.confirmationData.confirmationText === input.value) {
+      return this.errMessage = '';
+    }
+
+    this.errMessage = this.config.confirmationData.errorMessage;
+  }
```
galloween, line 50: this will happen on every keystroke.
i think it should be validated either only on blur - if `input.event === InputEvent.OnBlur`
+ it would be a good idea to disable the ok button if there is an error
+ you should be using only 1 validation method - the one in isValid()

galloween (changes requested):
I think it could be a little more organized, currently you have 2 places where you validate with 2 different methods.

if i was doing it, i would:
- validate on input blur (check inputEvents for type)
- disable/enable OK button and show/hide error message depending on validation 
- ok button just calls the action OR if the action is not a function or missing, returns whatever its supposed to return - Promise.resolve(true) probably

separately, i think everything should have fallbacks - i mean texts, so this should be expected to be usable without passing any config at all.

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
malkind: it won't work since it will just call itself over and over again. so this is why i'm saving the "original" CB function.
on the other hand, you suggested above to disable the ok button unless the text matches what we need, so if I do that, I  will not have to pass a different ok action to the dialog.

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
malkind: I don't like the onBlur option because this way you cannot type the work and immediately press "ok". you will have to click outside somewhere, let it update, and then click "ok".

plus I don't see any problem with the function being called on every keystroke, since there is no IO here.

projects/ui-framework/src/lib/popups/confirmation-dialog/confirmation-dialog.interface.ts
malkind: not sure how it's going to work
if they pass only confirmationText and pass neither label nor hintMessage, how should the user know what to type?

maybe I should make the label mandatory and all the rest optional.
@galloween

galloween: but it's supposed to be the same dialog most of the time, isnt it?
you need confirm you want to delete something.

correct me if I miss something.

it would be great to be able to use it without any config at all!

as for fallbacks, I think its easiest to put them in the template:
`[label]="config.confirmationData.label || ('some.default.translation' | translate)"` etc

galloween: or anywhere some prop from config is used:

`return this.bInput.value === (this.config.confirmationData.confirmationText || this.translate.instant('some.default.translation'))`

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
galloween: the only problem is it will be screaming the error in user's face before he finished typing...
so you need to at least debounce it..

by the way, it can be cool to make it all reactive:

```
valid$: Observable<boolean>;

this.valid$ = merge(
      this.bInput.changed.pipe(
        filter((event) => event.event !== InputEventType.onBlur),
        debounceTime(500)
      ),
      this.bInput.changed.pipe(
        filter((event) => event.event === InputEventType.onBlur)
      )
    ).pipe(
      map(
        (event) =>
          event.value ===
          (this.config.confirmationData?.confirmationText ||
            this.translate.instant('some.default.translation'))
      ),
      distinctUntilChanged()
    );
```
```
buttonConfig$: Observable<ConfirmationDialogButtons>

this.buttonConfig$ = valid$.pipe(
      startWith(false),
      map((valid) => {
        return {
          ...config.buttonConfig,
         cancel: {
           label: this.translate.instant('common.cancel'),
           ...this.config.buttonConfig.cancel,
          }
          ok: {
            label: this.translate.instant('common.ok'),
            ...this.config.buttonConfig.ok,
            disabled: !valid,
          },
        };
      })
    );
```
```
<b-dialog [dialogTitle]="config.title || ('default.title' | translate)"
          [dialogButtons]="buttonConfig$ | async">
  <div b-dialog-content>
    <p class="message" [innerHTML]="config.message || ('default.message' | translate)"></p>

    <b-input 
      [label]="config.confirmationData.label || ('default.label' | translate)"
      [placeholder]="config.confirmationData.placeholder || ('default.placeholder' | translate)"
      [hintMessage]="config.confirmationData.hintMessage || ('default.hint' | translate)"
      [errorMessage]="(valid$ | async) === false && (config.confirmationData.errorMessage || ('default.error' | translate))">
    </b-input>
  </div>
</b-dialog>
```

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  templateUrl: './delete-confirmation-dialog.component.html'
+})
+
+export class DeleteConfirmationDialogComponent implements OnInit, OnDestroy {
+  constructor(
+    public dialogRef: MatDialogRef<DeleteConfirmationDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public config: ConfirmationDialogConfig
+  ) {}
+
+  @ViewChild('bInput') bInput: InputComponent;
```
galloween, line 16: Viewchild can also select by the component/directive/whatever, and then you dont need the #name:

`@ViewChild(InputComponent) bInput: InputComponent;`

galloween (commented):
see comments - maybe you want to make it reactive?

i got carried away a bit with the code for it, sorry, 
but the key takeaway things there are:
- you can subscribe to input's change event from the viewchild
- you can split it be event type and debounce the typing, but not the blur

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,94 @@
...
+          cancel: {
+            label: this.translateService.instant('common.cancel'),
+            ...this.config.buttonConfig.cancel,
+          },
+          ok: {
+            label: this.translateService.instant('common.ok'),
+            ...this.config.buttonConfig.ok,
+            disabled: !valid,
+            type: ButtonType.negative
+          },
```
galloween, line 78: again, you need config?.buttonConfig?. etc

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  defaultLabel: string;
+  defaultConfirmationText: string;
+
+  private validitySubject: Subject<boolean> = new Subject();
+  private sub: Subscription;
+
+  ngOnInit() {
+    this.valid$ = this.validitySubject.asObservable();
+    this.setButtonsConfig();
+    this.setDefaultValues();
```
galloween, line 35: you can also move these to constructor body (so that values exist before first change detection, and they will not need to re-render after OnInit)

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  ngOnDestroy(): void {
+    this.dialogRef.close();
+    this.sub.unsubscribe();
+  }
+
+  ngAfterViewInit() {
+    this.sub = merge(
```
galloween, line 44: if you add static: true to the viewchild, the input will be available in ngOnInit

galloween (commented):
see comments

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export class DeleteConfirmationDialogComponent implements OnInit, OnDestroy, AfterViewInit {
+  constructor(
+    public dialogRef: MatDialogRef<DeleteConfirmationDialogComponent>,
+    @Inject(MAT_DIALOG_DATA) public config: ConfirmationDialogConfig,
+    private translateService: TranslateService
+  ) {}
+
+  @ViewChild(InputComponent) bInput: InputComponent;
+
```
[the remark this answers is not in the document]
malkind: thanks!

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,90 @@
...
+  valid$: Observable<boolean>;
+  buttonConfig$: Observable<ConfirmationDialogButtons>;
+
+  defaultLabel: string;
+  defaultConfirmationText: string;
+
+  private validitySubject: Subject<boolean> = new Subject();
```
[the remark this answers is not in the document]
malkind: removed the subject

projects/ui-framework/src/lib/popups/confirmation-dialog/confirmation-dialog.service.ts
```diff
@@ -18,4 +19,12 @@ export class ConfirmationDialogService {
       data: config,
     });
   }
+
+  openDeleteConfirmationDialog(config: ConfirmationDialogConfig): MatDialogRef<any> {
```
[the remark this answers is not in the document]
malkind: done

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.html
```diff
@@ -0,0 +1,13 @@
...
+    <p class="message" [innerHTML]="config.message || ('default.message' | translate)"></p>
+
+    <b-input
+      [label]="config.confirmationData.label || defaultLabel"
+      [placeholder]="config.confirmationData.placeholder || ('default.placeholder' | translate)"
+      [hintMessage]="config.confirmationData.hintMessage || ('default.hint' | translate)"
+      [errorMessage]="(valid$ | async) === false && (config.confirmationData.errorMessage || ('default.error' | translate))">
+    </b-input>
+  </div>
+</b-dialog>
```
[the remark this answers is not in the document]
malkind: done

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.defaultLabel = this.translateService.instant('bob-style.delete-confirmation.default-label', {
+      confirmationText: this.config.confirmationData.label ?
+        this.config.confirmationData.label :
+        this.defaultConfirmationText
+    });
+  }
+
+  ngOnDestroy(): void {
+    this.dialogRef.close();
+    this.sub.unsubscribe();
```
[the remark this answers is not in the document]
malkind: removed the subject. no more subscribe in the code.

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.setButtonsConfig();
+
+    this.defaultConfirmationText = this.config.confirmationData.confirmationText ?
+      this.config.confirmationData.confirmationText :
+      this.translateService.instant('common.delete');
+    this.defaultLabel = this.translateService.instant('bob-style.delete-confirmation.default-label', {
+      confirmationText: this.config.confirmationData.label ?
+        this.config.confirmationData.label :
+        this.defaultConfirmationText
+    });
```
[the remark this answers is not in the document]
malkind: why the "??"
config?property || default.value works.

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
```diff
@@ -0,0 +1,90 @@
...
+        debounceTime(500)
+      ),
+      this.bInput.changed.pipe(
+        filter((event) => event.event === InputEventType.onBlur)
+      )
+    ).pipe(
+      map(
+        (event) =>
+          event.value ===
+          (this.config.confirmationData?.confirmationText ||
```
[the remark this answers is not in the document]
malkind: nice!

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
malkind: ok

projects/ui-framework/src/lib/popups/confirmation-dialog/delete-confirmation-dialog/delete-confirmation-dialog.component.ts
malkind: thanks!!

galloween (approved):

PR #1836, opened 2020-11-19
removed empty state border

SkyWallkeRed:
list empty state got double border, removing it from parent MLAL


SkyWallkeRed:
fixed in previous pr.

PR #1843, opened 2020-11-23
table row drag

danidin:


projects/ui-framework/bob-table/src/table/table/styles/table-dnd.scss
```diff
@@ -0,0 +1,37 @@
...
+  ::ng-deep {
+    .ag-icon-grip {
+      display: block;
+      @include ag-icon('chevron-right');
+    }
+  }
+}
+
+:host-context(body) {
+
```
galloween, line 15: maybe better `::ng-deep` ? what value does scoping to body bring?

galloween (approved):

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -268,14 +272,48 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
...
+    const rowNeedsToMove = movingNode.id !== overNode.id;
+
+    if (rowNeedsToMove) {
+      const movingData = movingNode.data ;
+      const overData = overNode.data;
+      const fromIndex = this.rowData.indexOf(movingData);
+      const toIndex = this.rowData.indexOf(overData);
+      this.rowData.splice(fromIndex, 1);
+      this.rowData.splice(toIndex, 0, movingData);
+      this.gridApi.updateRowData({
```
galloween, line 295: better use `this.getGridApi()`
there are some edge cases where gridApi is not available. probably not in this case, but just for consistency getGridApi() may be better...

projects/ui-framework/bob-table/src/table/table/styles/table-dnd.scss
danidin: I'm actually trying to select `body .ag-dnd-ghost`, but `:host-context` only works for `body`, so I'm stuck

projects/ui-framework/bob-table/src/table/table/styles/table-dnd.scss
```diff
@@ -0,0 +1,23 @@
...
+  display: block;
+  @include ag-icon('drag');
+}
+
+::ng-deep .ag-dnd-ghost {
+  background: $white;
+  border-radius: 2px;
+  box-shadow: none;
+  padding: 4px;
+  border: 1px solid #424242;
```
galloween, line 15: maybe just $border-color ?

galloween (approved):

projects/ui-framework/bob-table/src/table/table/styles/table-dnd.scss
danidin: I don't know, this style is probably gonna change a lot

danidin: actually, why not

PR #1853, opened 2020-11-26
use getRowNodeId to compare nodes when dragging

danidin:


projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -282,9 +283,14 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
   }
 
   onRowDragMove(event: RowDragEvent) {
+    if (!isFunction(this.tableGridOptions.getRowNodeId)) {
+      return;
+    }
     const movingNode = event.node;
+    const movingNodeId = this.tableGridOptions.getRowNodeId(movingNode);
```
galloween, line 290: getRowNodeId has to be called with the (event.node.)**data**, not the node

galloween: but this is not how its supposed to work.

as i understand, you provide getRowNodeId in the options - and it will be used by the AgGrid to assign IDs to rows.
so you can keep using  overNode.id 
also, getRowNodeId is actually not required - if its absent, AgGrid will autogenerate ids.
more details here: https://www.ag-grid.com/javascript-grid-accessing-data/

galloween (changes requested):
see comments, i think your fist approach was right.
id is something agGrid adds automatically - either on its own, or using the provided (optional) getRowNodeId function

PR #1854, opened 2020-11-27
AutocompleteComponent. Eliminate redundant checks

MaxDsr:
Add a logic to check whether or not to openPanel on the options change inside ngOnChanges hook.
Check it inside the ngOnInit to escape further checks in ngOnChanges hook.

### DIFF
--- a/projects/ui-framework/src/lib/search/auto-complete/auto-complete.component.ts
+++ b/projects/ui-framework/src/lib/search/auto-complete/auto-complete.component.ts
@@ -81,20 +81,28 @@ export class AutoCompleteComponent implements OnInit, OnChanges, OnDestroy {
   private templatePortal: TemplatePortal;
   public panelOpen = false;
   private getFilteredOptions: () => AutoCompleteOption[] = this.skipFiltering;
+  private changeOptionsFlow: (changes: SimpleChanges) => void = this.changeOptions;
 
   ngOnInit(): void {
     this.getFilteredOptions = (this.skipOptionsFiltering) ? this.skipFiltering : this.filterOptions;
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/search/auto-complete/auto-complete.component.ts
```diff
@@ -81,22 +81,30 @@ export class AutoCompleteComponent implements OnInit, OnChanges, OnDestroy {
...
+  private changeOptions(changes: SimpleChanges) {
+    this.options = changes.options.currentValue;
+    this.filteredOptions = this.getFilteredOptions();
+  }
+
+  private changeOptionsAndOpenPanel(changes: SimpleChanges) {
+    this.changeOptions(changes);
+    this.openPanel();
+  }
+
```
galloween, line 107: not sure there is a real need to make code more complex :)
it was much more clear whats going on with the simple `if` in ngOnChanges

MaxDsr: Yes, but it actually goes every time through this `if` check on every options change. So I considered that it would be better somehow to escape this

galloween: well, in this case it's "much ado about nothing" :) and the readability suffers.
you have to look up and down the file to follow whats happening.

I personally would prefer less methods and the `this.skipOptionsFiltering ?` check both in getFilteredOptions and an in ngOnChanges (instead of switching methods).

also, just for info, assigning changes currentValue is there only for the test - because test call ngOnChanges manually.
In 'real life' by the time ngOnChanges happens, the options prop is already assigned.

but this is not that important :) we can merge.

galloween (approved):

PR #1866, opened 2020-12-06
Miki/change chevron to triangle

MikiAsh:
Replaced chevron 2 states icon in lists and collapsible section into a triangle variant

### DIFF
--- a/projects/ui-framework/package.json
+++ b/projects/ui-framework/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bob-style",
-  "version": "4.2.231",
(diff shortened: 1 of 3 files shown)


galloween:
you missed:
 - single select
 - single list
 - multi list

MikiAsh:
@galloween thanks for pointing out. single-select and single-list were added in https://github.com/hibobio/bob-style/pull/1868.
As for multi-list you had mentioned, I see it does have the new triangle icon.
I only replaced the icon for subtree expand, should I have also included the chevron down on dropdown-like components?

PR #1868, opened 2020-12-07
replaced icon in single-list, also affects single-select

MikiAsh:
Replace chevron icon to triangle in single-list.component.


MikiAsh:
@galloween 
I only replaced the icon for subtree expand, should I have also included the chevron down on dropdown-like components?

galloween:
> @galloween
> I only replaced the icon for subtree expand, should I have also included the chevron down on dropdown-like components?

no! you're right, the dropdowns are different and stay with chevrons

PR #1870, opened 2020-12-08
Add info-tooltip for label value component

MaxDsr:
Add info-tooltip for label value component.
Update story knobs.

asana task: https://app.asana.com/0/1159939249943132/1199081153131035/f
Screen Shot 2020-12-08 at 16 47 17

### DIFF
--- a/projects/ui-framework/src/lib/typography/label-value/label-value.component.scss
+++ b/projects/ui-framework/src/lib/typography/label-value/label-value.component.scss
@@ -135,6 +135,11 @@ $BLVTypeMap: (
   min-width: 0;
 }
 
+b-info-tooltip {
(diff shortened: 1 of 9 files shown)


projects/ui-framework/src/lib/typography/label-value/label-value.component.scss
```diff
@@ -135,6 +135,11 @@ $BLVTypeMap: (
   min-width: 0;
 }
 
+b-info-tooltip {
+  margin-left: times8(1);
+  line-height: normal;
```
galloween, line 140: if you put line-height: 0px, will be a liiiitle bit better

galloween (approved):
all good, see a couple small comments

PR #1871, opened 2020-12-08
table: add enableRowDrag input

danidin:


projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
```diff
@@ -42,16 +43,16 @@ export class TableUtilsService {
...
-    );
+  private getEnrichColumnDef(columnDefs: ColumnDef[], enableRowDrag: boolean): ColumnDef[] {
+    return columnDefs.map((colDef, i) => ({
+      ...colDef,
+      resizable: get(colDef, 'resizable', true),
+      sortable: get(colDef, 'sortable', true),
+      menuTabs: [],
+      cellClass: this.getCellClass(colDef),
+      cellStyle: this.getCellStyle(colDef),
+      rowDrag: enableRowDrag && i === 0
```
galloween, line 54: this will put false on all the other columns.
i suggest either putting true on all of them, or add this prop conditionally:
```
...
   cellStyle: this.getCellStyle(colDef),
   ...(enableRowDrag && i === 0 && {rowDrag:true})
```

galloween (approved):
see 1 comment about rowDrag prop on coldef

projects/ui-framework/bob-table/src/table/table-utils-service/table-utils.service.ts
danidin: I think it would be better to override any set rowDrag props for now, unless we support explicitly picking a column to display the drag handle on (right now the design is to always have it on the first column only)

galloween: ah it defines where the handle is! then ok!

projects/ui-framework/bob-table/src/table/table/table.component.ts
```diff
@@ -230,15 +230,13 @@ export class TableComponent extends AgGridWrapper implements OnInit, OnChanges {
...
-
-      this.rowDragEnabled = Boolean(this.columnDefs.find((def) => def.rowDrag));
     }
 
     if (hasChanges(changes, ['columnDefConfig'], true)) {
       previousColumnDefValue = changes.columnDefConfig.previousValue?.columnDef;
     }
 
-    if (hasChanges(changes, ['columnDefs', 'columnDefConfig'], true)) {
+    if (hasChanges(changes, ['columnDefs', 'columnDefConfig', 'enableRowDrag'], false)) {
```
danidin, line 239: @galloween check this out, I set discardTruthy do false here, so that it will restore sorting when dragging turned off

galloween: ok, so we still want the discardAllFalsey for columnDefs and columnDefConfig, so we can either:

1) split in 2 conditions: 
`hasChanges(changes, ['columnDefs', 'columnDefConfig'], true) || hasChanges(changes, ['enableRowDrag'], false)`
(this is easiest)

2) we can pass a truthyCheck function that will be used for the discardAllFalsey check, instead of the default Boolean:
`hasChanges(changes, ['columnDefs', 'columnDefConfig', 'enableRowDrag'], true, { truthyCheck: isValuevy })`
(isValuevy allows truthy, string, number, boolean, null)
or `truthyCheck: (val) => val !== undefined`

danidin: chose #2, thanks!

PR #1876, opened 2020-12-13
itable: add show menu on hover option

danidin:
part of https://app.asana.com/0/1163944084901313/1199534142933853/f

### DIFF
--- a/projects/ui-framework/bob-table/src/table/table/styles/table.component.scss
+++ b/projects/ui-framework/bob-table/src/table/table/styles/table.component.scss
@@ -307,6 +307,23 @@ $min-table-height: $table-header-height + $table-overlay-height + times8(3) * 2;
   }
 }
 
+[data-show-actions-on-hover='true'] ::ng-deep {
+
+  .ag-row {
+
+    .b-actions-cell {
+      display: none;
+    }
+
+    &.ag-row-hover {
+
+      .b-actions-cell {
+        display: flex;
+      }
+    }
+  }
+}
+
 [hidden] {
   display: block;
   visibility: hidden !important;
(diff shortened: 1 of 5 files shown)


projects/ui-framework/bob-table/src/table/table/styles/table.component.scss
```diff
@@ -307,6 +307,23 @@ $min-table-height: $table-header-height + $table-overlay-height + times8(3) * 2;
...
+    }
+
+    &.ag-row-hover {
+
+      .b-actions-cell {
+        display: flex;
+      }
+    }
+  }
+}
```
galloween, line 325: `b-actions-cell` is an element, not a class, so should be without the `.` dot

you can also use this 1 rule approach:
```
[data-show-actions-on-hover='true']::ng-deep .ag-row:not(.ag-row-hover) .b-actions-cell {
  display: none;
}
```
you need to try it, maybe having `display: none` will mess with the column width. in that case, you can use `visibility: hidden` or `opacity: 0`

projects/ui-framework/bob-table/src/table/table/table.component.html
```diff
@@ -13,6 +13,7 @@
                  (rowDragEnter)="rowIsDragged=true"
                  (rowDragEnd)="rowIsDragged=false; onRowDragEnd($event)"
                  [attr.data-row-hover]="styleConfig?.disableRowHoverBgColor ? false : true"
+                 [attr.data-show-actions-on-hover]="styleConfig?.showActionsOnHover"
```
galloween, line 16: because its a literal binding, its better to add `|| null` - this will remove the attribute completely for falsy showActionsOnHover, otherwise it will be `="undefined"`

galloween (commented):
see comments

projects/ui-framework/bob-table/src/table/table/styles/table.component.scss
danidin: it looks ok, `b-actions-cell` is a class name as well as the component selector

galloween: ah yes, it has a useless div inside with that class name

galloween (approved):

PR #1913, opened 2021-01-07
Change collapsible icon size to 32px

MikiAsh:
closes https://app.asana.com/0/1159939249943132/1199700964358533/f

### DIFF
--- a/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
+++ b/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
@@ -1,4 +1,5 @@
 @import '../../style/common-imports';
+@import '../../style/icons.mixin';
 
 $bcp-transition-length: 0.35s;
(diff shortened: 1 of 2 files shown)


projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
```diff
@@ -66,6 +67,7 @@ $bcp-header-min-height: 70px;
   min-height: $bcp-header-min-height;
   padding: times8(2) times8(4);
 
+  @include icon-size(32px);
```
galloween, line 70: @MikiAsh 
it will also look better if we keep the old width of the icon (so that it doesn use so much space), so lets add this:

```
@include icon-size(32px);

&:before {
  width: map-get($icon-sizes, large);  // this would be 24px
  display: flex;
  justify-content: center;
}

```

MikiAsh: done

galloween (approved):

PR #1976, opened 2021-03-02
Adjust color picker component

MaxDsr:
- add static width that is the same as min width for the select-panel in order to prevent bugs with color range selection
- add borders
- hide color type format selector
- include bob fonts
- add select for colors in the stories
- add color circle that displays selected color

Screen Shot 2021-03-02 at 15 30 22

asana : https://app.asana.com/0/0/1199985456001739/f


MaxDsr:
Merging for now to have some BETA in bob2 for this feature.

PR #2012, opened 2021-03-21
no auto close alert

doroncy:
added a flag to the alert config to disable the autohide of the alert


projects/ui-framework/src/lib/popups/alert/alert-service/alert.service.ts
```diff
@@ -27,6 +27,7 @@ export class AlertService {
   }
 
   public showAlert(config: AlertConfig): ComponentRef<AlertComponent> {
+    console.log('config', config);
```
galloween, line 30: isForgotToRemoveConsoleLog === true ? removeConsoleLog() : sleep()

galloween (approved):
isApproved = true

projects/ui-framework/src/lib/popups/alert/alert-service/alert.service.ts
doroncy: hehe

doroncy: we can rename to autoClose: boolean as well, what you think ?

PR #2018, opened 2021-03-29
bubble hover color

oriv:
add hover color input for bubble chart

### DIFF
--- a/projects/ui-framework/package.json
+++ b/projects/ui-framework/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bob-style",
-  "version": "4.3.114",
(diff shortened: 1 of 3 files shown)


projects/ui-framework/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "bob-style",
-  "version": "4.3.113",
+  "version": "4.3.114",
```
ronyrosenberg, line 3: I think this is done automatically once you merge to master

ronyrosenberg (approved):
looks good but @galloween knows better the area

PR #2020, opened 2021-03-29
Added linear shade utility to color.service

MikiAsh:


projects/ui-framework/src/lib/services/color-service/color.service.ts
```diff
@@ -72,4 +72,17 @@ export class ColorService {
       ? undefined
       : (color[0] * 299 + color[1] * 587 + color[2] * 114) / 1000;
   }
+
+    // https://stackoverflow.com/questions/5560248/programmatically-lighten-or-darken-a-hex-color-or-rgb-and-blend-colors
+    // color = rgb/rgba color
+    // offset = a number in the range of -1(=darkest) to 1(=brightest)
+    public LinearShade(color: string, offset: number) {
```
galloween, line 79: @MikiAsh 
1) its not a class, to be named with Capital letter :)
2) how about we add 2 more functions that normal people can use: darken & lighten. 
both should accept color string (hex or rgb, converted to rgb using existing method in same service) and a positive offset number (0-1)

PR #2026, opened 2021-04-01
Use $secondary css var in buttons

MikiAsh:
instead of using $grey

### DIFF
--- a/projects/ui-framework/src/lib/style/colors.scss
+++ b/projects/ui-framework/src/lib/style/colors.scss
@@ -14,9 +14,9 @@ $primary-500: #fea54a; // rgb(254,165,74)
 $primary-600: #ff962b; // rgb(255,150,43)
 $primary-700: #ff8100; // rgb(255,129,0)
 
-$secondary-500: #fecc4a; // rgb(254,204,74)
-$secondary-600: #f8bc20; // rgb(248,188,32)
-$secondary-700: #f0ad00; // rgb(240,173,0)
+$secondary-500: var(--secondary-500); // rgb(243,242,242)
+$secondary-600: var(--secondary-600); // rgb(226,226,226)
+$secondary-700: var(--secondary-700); // rgb(210,210,210)
 
 $positive-500: #1bd164; // rgb(27,209,100)
 $positive-600: #17b456; // rgb(23,180,86)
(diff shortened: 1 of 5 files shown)


projects/ui-framework/src/lib/style/colors.scss
```diff
@@ -14,9 +14,9 @@ $primary-500: #fea54a; // rgb(254,165,74)
 $primary-600: #ff962b; // rgb(255,150,43)
 $primary-700: #ff8100; // rgb(255,129,0)
 
-$secondary-500: #fecc4a; // rgb(254,204,74)
-$secondary-600: #f8bc20; // rgb(248,188,32)
-$secondary-700: #f0ad00; // rgb(240,173,0)
+$secondary-500: $grey-300;
+$secondary-600: $grey-400;
+$secondary-700: $grey-500;
```
galloween, line 19: no, we need something else:

**in lib/style/styles.css**

- change all `--primary-x` to use hex colors from the colors.scss:
```
--primary-500: #fea54a; // rgb(254,165,74)
```
- change all `--secondary-x` to use $grey-x:
```
--secondary-500: #{$grey-500};
```

**in lib/style/colors.css**
- change all $primary-x and $secondary-x to use css variables:
```
$primary-500: var(--primary-500);
$secondary-500: var(--secondary-500);
```

projects/ui-framework/src/lib/style/component-mixins/button.mixin.scss
```diff
@@ -55,13 +55,13 @@ $button-size-large: 44px;
     }
 
     @if $type == 'secondary' {
-      background-color: $grey-400;
+      background-color: $secondary-600;
```
galloween, line 58: to keep in line with other styles in this file, use `var(--secondary-xxx)`

projects/ui-framework/src/lib/style/colors.scss
galloween: actually this change will have broader effect than just buttons.

if you want to scope this to buttons, 
only do this change:
```
--secondary-XXX: #{$grey-XXX};
```

galloween (commented):
>>>

galloween (approved):

PR #2039, opened 2021-04-07
Rremove old functionality

MaxDsr:


projects/ui-framework/src/lib/form-elements/color-picker/color-picker.component.ts
```diff
@@ -76,28 +76,16 @@ export class ColorPickerComponent extends BaseFormElement implements OnDestroy,
     const value = event.target.value;
 
     // tslint:disable-next-line: triple-equals
-    if (value != this.value && this.isValidValue(value)) {
-      this.errorMessage = '';
+    if (value != this.value) {
       this.writeValue(value, this.forceElementValue);
       this.transmitValue(this.value, {
         eventType: [InputEventType.onChange]
```
galloween, line 82: this should now be InputEventType.onBlur, but not critical

PR #2047, opened 2021-04-18
prevent order in editable list

micaelpc:
this PR is follow up to the asana task:
Screen Recording 2021-04-19 at 8 56 46

@galloween let me know if it's ok so i can move on to unit tests.

relates: https://app.asana.com/0/1159573723793308/1200183470801918/f

### DIFF
--- a/projects/ui-framework/src/lib/lists/editable-list/editable-list.component.html
+++ b/projects/ui-framework/src/lib/lists/editable-list/editable-list.component.html
@@ -82,18 +82,20 @@
                       [lockAxis]="'y'"
                       [dragClass]="'bel-item-dragged'"
                       [dropClass]="'bel-item-dropped'"
-                      [nonDragAreaSelector]="'.bel-item-confirm, .bel-trash-button'"
+                      [nonDragAreaSelector]="'.bel-item-confirm, .bel-trash-button, .non-draggable-item'"
                       (dragStart)="onDragStart()"
                       (drop)="onDrop($event)">
 
   <smooth-dnd-draggable *ngFor="let item of listState.list; let index = index; trackBy: listTrackBy"
                         role="listitem">
 
-    <div class="bel-item b-icon-drag-alt b-icon-medium b-icon-dark"
+    <div class="bel-item b-icon-medium b-icon-dark"
          [attr.data-item-id]="item.id"
          [ngClass]="{'new': addedItem && index === 0,
                      'deleted':removedItem && removingIndex === index,
-                     'warn': sameItemIndex === index}">
+                     'warn': sameItemIndex === index,
+                     'b-icon-drag-alt': !allowedActions.preventOrder,
+                     'non-draggable-item': allowedActions.preventOrder }">
 
       <span class="bel-item-value"
             data-max-lines="2">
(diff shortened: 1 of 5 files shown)


projects/ui-framework/src/lib/lists/editable-list/editable-list.stories.ts
```diff
@@ -114,6 +115,7 @@ story.add(
         allowSort: boolean('allowSort', true, 'Props'),
         allowAdd: boolean('allowAdd', true, 'Props'),
         allowRemove: boolean('allowRemove', true, 'Props'),
+        preventOrder: boolean('preventOrder', false, 'Props'),
```
omer-r, line 118: i'd definitely follow the pattern and call it `allowOrder` instead of prevent like all other attributes

micaelpc: @omer-r - agree with you.
the only problem is that this is part of a whole object as Input. so my concern is to make a regression for other usages of the "editable-list". so if we take the right way, I need first to:
1.change the interface to : 
`export interface EditableListActions {
    sort?: boolean;
    add?: boolean | string;
    remove?: boolean;
    order?: boolean;
}`
2. change the 4 usages in bob:
image

image
3. make the change in bob-style.

@omer-r @galloween that's plan is ok?

galloween: you dont have to change existing uses, just use  `allowedActions.order === false` (for no order) and/or `allowedActions.order !== false` (for order allowed)

projects/ui-framework/src/lib/lists/editable-list/editable-list.component.scss
```diff
@@ -94,6 +94,9 @@ $bel-item-dragged-border-color: $grey-400;
 
   cursor: move;
   cursor: grab;
+  .non-draggable-item {
+    cursor: default !important;
+  }
```
galloween, line 99: this should be  `&.non-draggable-item` and then the important is not needed

projects/ui-framework/src/lib/lists/editable-list/editable-list.component.html
```diff
@@ -82,18 +82,20 @@
                       [lockAxis]="'y'"
                       [dragClass]="'bel-item-dragged'"
                       [dropClass]="'bel-item-dropped'"
-                      [nonDragAreaSelector]="'.bel-item-confirm, .bel-trash-button'"
+                      [nonDragAreaSelector]="'.bel-item-confirm, .bel-trash-button, .non-draggable-item'"
```
galloween, line 85: just for documentation, i found another way to disable the drag-n-drop - to give it bad selector for drag handle:

[dragHandleSelector]="allowedActions.order===false ? 'non-existing-element' : null"

galloween (commented):
see comment about how to go about 'truthy undefined' property

projects/ui-framework/src/lib/lists/editable-list/editable-list.stories.ts
micaelpc: @galloween great idea!
DONE

projects/ui-framework/src/lib/lists/editable-list/editable-list.component.scss
micaelpc: DONE

galloween (approved):

PR #2056, opened 2021-04-21
add label-secondary type to chips

talitamar91:
closes https://app.asana.com/0/1159939249943132/1199935973666933/f


galloween (approved):

projects/ui-framework/src/lib/chips/chip/chip.component.scss
```diff
@@ -150,12 +150,19 @@ $chipTypeMap: (
...
       border-radius: $border-radius/2,
       height: times8(2),
       padding: 0 8px,
     ),
+  ),
+  label-secondary: (
+    '*': (
+      background-color: white,
+      color: $grey-600,
+      border-color: $grey-500,
```
galloween, line 164: all border colors should use $border-color (which is same as $grey-500)

PR #2063, opened 2021-04-25
use hard-coded 'DELETE' string from ts file

MikiAsh:
relates https://app.asana.com/0/1165108539203930/1200135502243107/f


doroncy (approved):

doroncy (approved):

galloween (changes requested):
it already supports inputting 'delete'
a better update would be making sure it uses the same string in the text and in the check.
(its ok to ask for 'Radera', but it should also check for 'Radera')

PR #2076, opened 2021-04-28
add defualt empty string to localization

talitamar91:
add defualt empty string to localization in card-table and single select

### DIFF
--- a/projects/ui-framework/src/lib/table/card-table/card-table/card-table.component.html
+++ b/projects/ui-framework/src/lib/table/card-table/card-table/card-table.component.html
@@ -31,8 +31,8 @@
 </ul>
 
 <div class="card-table-body"
-     *ngIf="default && !(table?.length)">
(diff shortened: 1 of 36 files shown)


galloween (approved):

projects/ui-framework/src/lib/table/card-table/card-table/card-table.component.html
```diff
@@ -31,8 +31,8 @@
 </ul>
 
 <div class="card-table-body"
-     *ngIf="default && !(table?.length)">
+     *ngIf="!(table?.length)">
   <p class="table-card-empty b-display-3">
-    {{ default }}
+    {{ default || 'bob-style.card-table.default-empty' | translate}}
   </p>
```
galloween, line 37: what if the consumer does not want to show the empty state?

how about 
```
*ngIf="default !== false && !(table?.length)"
```

PR #2077, opened 2021-04-28
List group selectable

maximblack:
related to: https://app.asana.com/0/1159966416653227/1200242356908371/f

### DIFF
--- a/projects/ui-framework/src/lib/lists/list-service/list-model.service.ts
+++ b/projects/ui-framework/src/lib/lists/list-service/list-model.service.ts
@@ -86,9 +86,7 @@ export class ListModelService {
             : selectedCount === group.options.length,
           indeterminate: this.isIndeterminate(group.options, selectedCount),
           selectedCount,
-          hasCheckbox: isBoolean(group.groupSelectable)
-            ? group.groupSelectable && hasCheckbox
-            : hasCheckbox,
+          hasCheckbox: hasCheckbox && group.groupSelectable !== false,
           groupIsOption,
         };
       }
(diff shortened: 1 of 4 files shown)


projects/ui-framework/src/lib/lists/list.interface.ts
```diff
@@ -8,6 +8,7 @@ export type itemID = string | number;
 export interface SelectGroupOption {
   groupName: string;
   groupIndex?: number;
+  groupSelectable?: boolean;
```
galloween, line 11: can you also add it to interface doc here:
projects/ui-framework/src/lib/lists/list.interface.md (its on top of the file)

projects/ui-framework/src/lib/lists/list-service/list-model.service.ts
```diff
@@ -86,7 +86,9 @@ export class ListModelService {
             : selectedCount === group.options.length,
           indeterminate: this.isIndeterminate(group.options, selectedCount),
           selectedCount,
-          hasCheckbox,
+          hasCheckbox: isBoolean(group.groupSelectable)
+            ? group.groupSelectable && hasCheckbox
+            : hasCheckbox,
```
galloween, line 91: its better to add the check here:
projects/ui-framework/src/lib/lists/list-element.abstract.ts  line 542 (because thats where other conditions for checkbox are)
also, you can just add:
```
...  && group.groupSelectable !== false
```
this will allow for "truthy undefined"

galloween (approved):

PR #2082, opened 2021-05-03
add edit functionality each item in editbale-list component

talitamar91:
closes https://app.asana.com/0/1159939249943132/1200004352134432/f
add edit action to allowedActions object and  enable edit each item in the list

### DIFF
--- a/projects/ui-framework/src/lib/lists/editable-list/editable-list.component.html
+++ b/projects/ui-framework/src/lib/lists/editable-list/editable-list.component.html
@@ -29,7 +29,8 @@
         <b-button class="bel-cancel-button"
                   [type]="buttonType.tertiary"
                   [size]="buttonSize.small"
-                  [text]="('common.cancel' | translate)">
+                  [text]="('common.cancel' | translate)"
+                  (clicked)="resetActionsData()">
         </b-button>
         <b-button class="bel-done-button"
                   [type]="buttonType.secondary"
@@ -42,13 +43,9 @@
     </div>
   </div>
 
-  <p b-input-message
-     *ngIf="addingItemLen"
-     [hidden]="!addingItem || !(inputInvalid || maxChars - addingItemLen < 15)"
-     [errorMessage]="inputInvalid ? '&#34;' + listState.list[sameItemIndex].value + '&#34; ' + ('list-editor.editor.already-exists' | translate) : null"
-     [maxChars]="maxChars - addingItemLen < 15 ? maxChars : null"
-     [length]="addingItemLen">
-  </p>
+    <ng-container 
+    *ngTemplateOutlet="errorMsg; context: {'showError': addingItemLen ,
+     'hidden': !addingItem || !(inputInvalid || maxChars - addingItemLen < 15)}"></ng-container>
 
   <div class="bel-header-bottom">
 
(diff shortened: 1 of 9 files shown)


projects/ui-framework/src/lib/lists/editable-list/editable-list.component.ts
```diff
@@ -165,6 +162,77 @@ export class EditableListComponent implements OnChanges, OnInit, OnDestroy {
...
+          this.editInputInvalid = true;
+          return;
+      }
+      this.listState.list[this.removingIndex].value = this.currentEditInput.nativeElement.value
+      this.removingIndex = null;
+      this.transmit();
+    } else {
+      this.removingIndex = null;
+    }
+  }
```
galloween, line 216: the edit functionality should be simplified. 
here is the generic idea:

```
editItem(item: SelectOption, index: number): void {
    this.removingIndex = index; // better rename to something more appropriate
    item.originalValue = item.value; // copy value to be able to reset
   ...
}

<input #editItemInput [(ngModel)]="item.value" ...

<b-button (clicked)="onEditCancel(item, editItemInput)" ...> Cancel
<b-button (clicked)="onEditSave(item)" ...> Save

onEditCancel(item: SelectOption, input: HTMLInputElement) {
  input.value = item.value = item.originalValue; // reset value
  delete item.originalValue; // cleanup
 ...
}

onEditSave(item: SelectOption) {
    delete item.originalValue; // only cleanup needed, value already in sync
  ...
}
```

projects/ui-framework/src/lib/lists/editable-list/editable-list.enum.ts
```diff
@@ -3,3 +3,9 @@ export enum ListSortType {
   Asc = 'Asc',
   Desc = 'Desc',
 }
+
+
+export enum IconActionsType {
+  Delete = 'Delete',
+  Rename = 'Rename'
+}
```
galloween, line 11: i dont think we need this enum:
- its confusing that the actions are named differently from the EditableListActions interface
- as it is only for internal use, you can just use strings, which you can typecheck like this:
```
type IconActions = 'remove' | 'edit';

someMethod(action: IconActions) { ... }

someMethod('remove') // will be typechecked, only 'rename' and 'delete' allowed
```

if you need the type some object keys with it:
```
const actionsDeafults: Record<IconActions, boolean> = {
  remove: true,
  edit: false,
  jump: 1 // not allowed
}
```

galloween (changes requested):
i started to comment, but the code is too big to effectively review it...

basically, we need to simplify everything 👨‍🏫

- for listState, leave only **list**; get rid of **all** the other properties, noone uses them
- remove all the logic about comparing item names - same names are now allowed
- for edit, bind the input directly to the list item, no need for complex logic 
- bring back the 'cancel on blur' functionality
- make sure only 1 edit or add input is enabled at any given time

projects/ui-framework/src/lib/lists/editable-list/editable-list.component.html
```diff
@@ -42,13 +43,9 @@
...
-  <p b-input-message
-     *ngIf="addingItemLen"
-     [hidden]="!addingItem || !(inputInvalid || maxChars - addingItemLen < 15)"
-     [errorMessage]="inputInvalid ? '&#34;' + listState.list[sameItemIndex].value + '&#34; ' + ('list-editor.editor.already-exists' | translate) : null"
-     [maxChars]="maxChars - addingItemLen < 15 ? maxChars : null"
-     [length]="addingItemLen">
-  </p>
+    <ng-container 
+    *ngTemplateOutlet="errorMsg; context: {'showError': addingItemLen ,
+     'hidden': !addingItem || !(inputInvalid || maxChars - addingItemLen < 15)}"></ng-container>
```
galloween, line 48: nice touch - to not repeat the template

galloween (commented):
i got carried away a little bit and i ended up doing more cleanup/refactoring....
its pretty much completely different now, but hopefully a little less messy...

so now its collaborative effort! 👬

PR #2106, opened 2021-05-11
David/summary insights

SkyWallkeRed:
Screen Shot 2021-05-23 at 17 08 54
todo: story


projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.html
```diff
@@ -0,0 +1,39 @@
+<div class="inl-blk" *ngFor="let item of data; let last = last" [class.brd-r]="!last" >
+
+  <div class="insight-pad">
```
galloween, line 3: you dont need this wrapper.
put padding classes directly in b-label-value: 
`pad-l-16 pad-r-40` (or whatever)
but 1st element probably should not have left padding, and last  element - right.
(you can also use `first` from ngFor)

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.html
```diff
@@ -0,0 +1,39 @@
...
+    </b-label-value>
+  </div>
+
+  <div class="insight-pad" *ngIf="item.type === summaryInsightType.progressBar">
+    <b-progress-bar class="inl-blk"
+                    [type]="item.data.type"
+                    [size]="item.data.size"
+                    [data]="item.data.data"
+                    [config]="item.data.config">
+      <span class="progress-bar-inner-label" header-left>{{item.label.value}}</span>
```
galloween, line 35: dont think you need this element: 
the data already supports 2 text inputs for the header - primary (left side) and secondary (right side)
 (their css is the same)

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.html
```diff
@@ -0,0 +1,39 @@
...
+    <b-progress-donut class="inl-blk progress-donut"
+                      [size]="item.data.size"
+                      [donutSize]="item.data.donutSize"
+                      [data]="item.data.data"
+                      [config]="item.data.config">
+    </b-progress-donut>
+    <b-label-value class="inl-blk pad-l-16"
+                   [labelValue]="item.label">
+    </b-label-value>
+  </div>
```
galloween, line 27: dont think this type (progressDonutWithLabelValue) is needed - donut has built-in text

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.html
```diff
@@ -0,0 +1,39 @@
+<div class="inl-blk" *ngFor="let item of data; let last = last" [class.brd-r]="!last" >
```
galloween, line 1: it would be better if for ngFor you use ng-container,
and move the brd-r to the components (b-label-value, b-progress-x), so that all layout classes are in the same place (border, padding)

galloween: also inl-blk is totally not needed, as they are in a flex container

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.html
```diff
@@ -0,0 +1,39 @@
...
+                      [data]="item.data.data"
+                      [config]="item.data.config">
+    </b-progress-donut>
+    <b-label-value class="inl-blk pad-l-16"
+                   [labelValue]="item.label">
+    </b-label-value>
+  </div>
+
+  <div class="insight-pad" *ngIf="item.type === summaryInsightType.progressBar">
+    <b-progress-bar class="inl-blk"
```
galloween, line 30: get rid of insight-pad (class and div)
put ngIf and padding classes directly on b-progress-bar
also remove inl-blk

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../style/common-imports';
+@import '../../style/icons.mixin';
+
+:host {
+  display: flex;
+  align-content: flex-start;
+  width: 100%;
```
galloween, line 7: are you sure you want to align the components to top?

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.scss
```diff
@@ -0,0 +1,23 @@
...
+
+  .insight-pad {
+    padding-left: times8(2);
+    padding-right: times8(5);
+  }
+
+  .progress-bar-inner-label{
+    color: $grey-600;
+  }
```
galloween, line 16: these 2 classes will not be necessary

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.scss
```diff
@@ -0,0 +1,23 @@
...
+
+  ::ng-deep {
+    .blv-value {
+      height: 100%;
+    }
+  }
```
galloween, line 22: this doesnt look right. and probably does nothing.
what did you try to achieve?

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.interfaces.ts
```diff
@@ -0,0 +1,8 @@
+import { LabelValue, ProgressBar, ProgressDonut } from 'bob-style';
+import { SummaryInsightType } from './summary-insights.enums';
+
+export interface SummaryInsight {
+  type: SummaryInsightType,
+  label?: LabelValue,
+  data?: (ProgressBar | ProgressDonut)
+}
```
galloween, line 8: if we get rid of progressDonutWithLabelValue then
 `data` can be `LabelValue | ProgressBar | ProgressDonut`

projects/ui-framework/src/lib/layout/summary-insights/summary-insights.component.html
SkyWallkeRed: the color of the text is not as requested :/

galloween (approved):

projects/ui-framework/src/lib/indicators/progress/progress-bar/progress-bar.component.scss
```diff
@@ -90,9 +90,10 @@ $BPBtypeMap: (
 );
 
 :host {
-  @include progress-host('bpb', grid);
+  @include progress-host('bpb', $display: grid, $width: null);
   text-align: left;
-  min-width: 0;
+  min-width: 120px;
```
omerhochbob, line 95: @SkyWallkeRed @galloween 
This line causes a regression for us in some places.
Is this a must as part of the progress bar?

Or do you have an option to revert back to min-width: 0? And change it in your usage?

PR #2108, opened 2021-05-11
Created a 1-third-2-thirds layout component

MikiAsh:
relates https://app.asana.com/0/1200306097679416/1200306097679429/f


projects/ui-framework/src/lib/layout/third-2thirds/third-2thirds.component.scss
```diff
@@ -0,0 +1,28 @@
...
+    display: flex;
+    flex-direction: column;
+  }
+  [side] {
+    flex: none;
+    width: 384px;
+    display: flex;
+    flex-direction: column;
+  }
+}
```
galloween, line 28: 1)
the issue here is that this will not work:
```
<ng-container main> hello </ng-container>
```
(its valid for ng-content, but does not create an element with attribute [main]).
in your component you should style elements that are directly in your template

2)
the margin should be only between the columns

3)
you missed the max width 1200px on the container.

---------------------

better approach:
```
:host {
  display: grid;
  grid-template-columns: 384px 1fr;
  grid-column-gap: 24px;
  max-width: 1200px;
}
```
and for template - use `''` (empty string). thats it.

not sure this even needs to be a component - you can just create it as global style:

```
.b-third-2thirds, [b-third-2thirds] {
  display: grid;
  grid-template-columns: 384px 1fr;
  grid-column-gap: 24px;
  max-width: 1200px;
}
```
now you can add this layout to anything with class or attribute

PR #2111, opened 2021-05-12
Select and view - create the left list select

rinat-shemer:
added only the left side of the select and view - the select list
when clicking on an item, it disappears from the list (by filtering the selected items from the original list)
no css yet
fix startWithGroupsCollapsed changes bug - when there is no change and the list is not empty it should stay as it was initialised 

Screen Shot 2021-05-12 at 14 19 28

-------------------------------------------------------------------------------------------------------------------------

cannot merge to master so I added the right side as well
to dos:
design issues (put them side by side, height)
add empty state for the view list
serach
drag and drop

Screen Shot 2021-05-18 at 17 23 24


projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,75 @@
...
+  @InputObservable([])
+  @Input('options')
+  public inputOptions$: Observable<SelectGroupOption[]>;
+
+  @InputObservable(null)
+  @Input('value')
+  public inputValue$: Observable<itemID[]>;
+
+  public selectOptions$: BehaviorSubject<SelectGroupOption[]> = new BehaviorSubject(undefined);
+  public listValue$: BehaviorSubject<itemID[]> = new BehaviorSubject(null);
```
galloween, line 27: you can use `InputSubject`, then you will not need separate listValue$:

```
@InputSubject([]) @Input('value') public listValue$: BehaviorSubject<itemID[]>;
```

now you will be able to push to it

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,75 @@
...
+    unsubscribeArray(this.subs);
+
+    [
+      this.selectOptions$,
+    ].forEach((subj) => subj.complete());
+  }
+
+  private getOptionsWithoutSelected(options: SelectGroupOption[], value: itemID[]): SelectGroupOption[] {
+    return chain(options)
+      .map(group => ({
```
galloween, line 59: lets get rid of lodash please? the main code will look exactly the same:

<s>chain(options)</s> options
&nbsp;&nbsp;.map(...)
&nbsp;&nbsp;.filter(...)
&nbsp;&nbsp;<s>.value()</s>

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,75 @@
...
+      .filter(group => group.options.length)
+      .value();
+  }
+
+  public onSelectChange(listChange: ListChange): void {
+    const selected = listChange.getSelectedIds();
+    const currentValue = this.listValue$.getValue();
+
+    this.listValue$.next(currentValue.concat(selected));
+  }
```
galloween, line 74: we can also automate it:
```
@ViewChild(SingleListComponent, {static: true}) listComp: SingleListComponent;

ngOnInit() {

   this.listComp.selectChange.pipe(
      map( (listChange) =>  asArray(this.listValue$.getValue()).concat( listChange.selectedIDs || [] )  )
   )
  .subscribe( this.listValue$  )

}
```

the asArray and ||[] will safeguard against adding null to the array.

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,75 @@
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class SelectAndViewComponent {
+export class SelectAndViewComponent implements OnInit, OnDestroy {
+
+  @InputObservable([])
+  @Input('options')
+  public inputOptions$: Observable<SelectGroupOption[]>;
+
+  @InputObservable(null)
```
galloween, line 22: i think its safer to always keep it an array, so start with []

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,75 @@
...
+
+    this.subs.push(
+      combineLatest([
+        this.inputOptions$,
+        this.listValue$,
+      ])
+        .pipe(
+          map(([options, value]) => isArray(value)
+            ? this.getOptionsWithoutSelected(options, value)
+            : options),
```
galloween, line 43: change to:  `isNotEmptyArray(value)` - this will return false for both null and []

projects/ui-framework/src/lib/lists/list-element.abstract.ts
```diff
@@ -199,12 +200,9 @@ export abstract class BaseListElement
...
-    }
-
-    if (hasChanges(changes, ['startWithGroupsCollapsed', 'options'])) {
-      this.startWithGroupsCollapsed =
-        (this.startWithGroupsCollapsed || !changes.startWithGroupsCollapsed) &&
-        isNotEmptyArray(this.options, 1);
+      this.startWithGroupsCollapsed = isEmptyArray(this.options)
+        ? false
+        : this.startWithGroupsCollapsed;
     }
```
galloween, line 206: this is a **very** fragile place to make changes. 
startWithGroupsCollapsed also affects some later choices in ngOnChanges.
there is some particular reason why it is like this, but i dont remember. i only remember it was revisited multiple times already....

what did you try to achieve/solve?

galloween (commented):
great!!! like it a lot.
see comments - it can be a optimised even a little more!

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,81 @@
...
+      }))
+      .filter(group => group.options.length);
+  }
+
+  public onSelectChange(listChange: ListChange): void {
+    const selected = listChange.getSelectedIds();
+    const currentValue = this.listValue$.getValue();
+
+    this.listValue$.next(currentValue.concat(selected));
+  }
```
galloween, line 80: so now this method is not needed

galloween (commented):
👍🏻

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -1,10 +1,81 @@
...
+
+    this.subs.push(
+      this.listComponent.selectChange
+        .pipe(
+          map((listChange) =>  asArray(this.listValue$.getValue())
+            .concat(listChange.selectedIDs || []))
+        )
+        .subscribe(this.listValue$)
+    );
+  }
```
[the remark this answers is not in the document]
galloween: you can also push to subs in one go:
```
this.subs.push( 
  a$.subscribe(..),  
  b$.subscribe()  
)
```

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.html
```diff
@@ -1,3 +1,18 @@
 <b-single-list [options]="selectOptions$ | async"
                [startWithGroupsCollapsed]="false">
 </b-single-list>
+
+<div *ngIf="viewList$ | async as viewList"
+     class="viewList">
```
galloween, line 6: - there is no benefit for this ngIf, you can ngFor on async, it will show nothing until it has value, by itself
- need trackBy!

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -0,0 +1,29 @@
...
+.viewList {
+  border-radius: $border-radius;
+  border: 1px solid $border-color;
+  box-shadow: 0 2px 3px 0 $boxShadowColor;
+  overflow: hidden;
+}
+
+.viewItem {
+  grid-template-columns: 40px auto 40px;
+  height: times8(5.5);
```
galloween, line 12: - if its not a proper 8 multiplier, its easier to just put pixel value - at least then you dont need calculator to understand it
- actually we have a variable for 44px, `$wrap-height`

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -0,0 +1,29 @@
+@import '../../style/common-imports';
+
+.viewList {
+  border-radius: $border-radius;
+  border: 1px solid $border-color;
+  box-shadow: 0 2px 3px 0 $boxShadowColor;
```
galloween, line 6: `<div class="brd rounded shadow">`

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.viewItem {
+  grid-template-columns: 40px auto 40px;
+  height: times8(5.5);
+  text-align: left;
+  border-bottom: 1px solid $border-color;
+
+  &:hover,
+  &.focus {
```
galloween, line 17: you are not using this class...

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -68,4 +98,24 @@ export class SelectAndViewComponent implements OnInit, OnDestroy {
...
+
+  private getViewListData(options: SelectGroupOption[], value: itemID[]): ViewListItem[] {
+    const data = [];
+
+    options.forEach(group =>
+      group.options.forEach(option =>
+        asArray(value).includes(option.id)
+        && data.push({ value: option.value, groupName: group.groupName, id: option.id })));
+
+    return data.sort((a, b) => value.indexOf(a.id) < value.indexOf(b.id) ? -1 : 1);
```
galloween, line 110: why value would contain id ?

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -68,4 +98,24 @@ export class SelectAndViewComponent implements OnInit, OnDestroy {
...
+  }
+
+  public removeItemFromList(itemId: string): void {
+    const currentValue = this.listValue$.getValue();
+    const index = currentValue.indexOf(itemId);
+
+    currentValue.splice(index, 1);
```
galloween, line 117: there is a helper function that does same thing + add slice (because splice mutates array)
```
 this.listValue$.next(
      arrayRemoveItemMutate( this.listValue$.getValue()?.slice(), itemId )
 )
```

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -68,4 +98,24 @@ export class SelectAndViewComponent implements OnInit, OnDestroy {
...
+
+    options.forEach(group =>
+      group.options.forEach(option =>
+        asArray(value).includes(option.id)
+        && data.push({ value: option.value, groupName: group.groupName, id: option.id })));
+
+    return data.sort((a, b) => value.indexOf(a.id) < value.indexOf(b.id) ? -1 : 1);
+  }
+
+  public removeItemFromList(itemId: string): void {
```
galloween, line 113: maybe stay with `itemID` type....

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
```diff
@@ -68,4 +98,24 @@ export class SelectAndViewComponent implements OnInit, OnDestroy {
...
       .filter(group => group.options.length);
   }
+
+  private getViewListData(options: SelectGroupOption[], value: itemID[]): ViewListItem[] {
+    const data = [];
+
+    options.forEach(group =>
+      group.options.forEach(option =>
+        asArray(value).includes(option.id)
+        && data.push({ value: option.value, groupName: group.groupName, id: option.id })));
```
galloween, line 108: why bother? just push the `option`.
with remapping you just use more memory (vs. passing reference to already stored object)
its also 3 times faster (to pass existing option), but of course in the sub-millisecond range it doesn't matter...

galloween (commented):
yes! can be optimised just a little bit

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -0,0 +1,29 @@
...
+  border: 1px solid $border-color;
+  box-shadow: 0 2px 3px 0 $boxShadowColor;
+  overflow: hidden;
+}
+
+.viewItem {
+  grid-template-columns: 40px auto 40px;
+  height: times8(5.5);
+  text-align: left;
+  border-bottom: 1px solid $border-color;
```
galloween, line 14: you could also use class for the border: `brd-b`

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.html
rinat-shemer: I need to add empty state in case of empty array, so I will need the `ngIf`

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
rinat-shemer: need it for the remove function

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
rinat-shemer: the option does not contain the group name and I need it for the display

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.html
rinat-shemer: added the track by

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -0,0 +1,26 @@
+@import '../../style/common-imports';
+
+.viewList {
+  border-radius: $border-radius;
+  border: 1px solid $border-color;
```
galloween, line 5: so we dont need these 2 any more

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -0,0 +1,26 @@
...
+  overflow: hidden;
```
galloween, line 6: we probably dont need this one as well :)

you can do this so that text doesnt wrap:
```
<div data-max-lines="1">{{item.value}}</div> // other numbers will also work
```

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.ts
galloween: ok then you should do:
```
=> value.indexOf(a.id) - value.indexOf(b.id)
```
(because in your code you miss the case where they are equal and 0 should be returned)

projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.html
```diff
@@ -0,0 +1,18 @@
+<b-single-list [options]="selectOptions$ | async"
+               [startWithGroupsCollapsed]="false">
+</b-single-list>
+
+<div *ngIf="viewList$ | async as viewList"
+     class="viewList brd rounded shadow">
```
galloween, line 6: you can add class `scroll-visible-alt`
so that it has same scroll style as the list

PR #2121, opened 2021-05-17
Added tooltip for menu items(b-menu)

igorCuznetov:
Added tooltip for menu items(b-menu)

## Asana
https://app.asana.com/0/1198038441420635/1200010090826472/f

## Screenshot
Capture

### DIFF
--- a/projects/ui-framework/src/lib/navigation/menu/menu.component.html
+++ b/projects/ui-framework/src/lib/navigation/menu/menu.component.html
@@ -61,7 +61,8 @@
                 (click)="onClick(item)"
                 [disabled]="itemIsDisabled(item)"
                 [disableRipple]="true">
-          <span [matTooltip]="item.tooltip">{{item.label}}</span>
+          <span *ngIf="!item.tooltip">{{item.label}}</span>
+          <span *ngIf="item.tooltip" class="flx-grow" [matTooltip]="item.tooltip">{{item.label}}</span>
         </button>
       </ng-container>
 
--- a/projects/ui-framework/src/lib/navigation/menu/menu.component.html
+++ b/projects/ui-framework/src/lib/navigation/menu/menu.component.html
@@ -61,7 +61,7 @@
                 (click)="onClick(item)"
                 [disabled]="itemIsDisabled(item)"
                 [disableRipple]="true">
-          <span>{{item.label}}</span>
+          <span [matTooltip]="item.tooltip">{{item.label}}</span>
         </button>
       </ng-container>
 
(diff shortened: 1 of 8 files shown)


projects/ui-framework/src/lib/navigation/menu/menu.component.html
```diff
@@ -61,7 +61,7 @@
                 (click)="onClick(item)"
                 [disabled]="itemIsDisabled(item)"
                 [disableRipple]="true">
-          <span>{{item.label}}</span>
+          <span [matTooltip]="item.tooltip">{{item.label}}</span>
```
galloween, line 64: add class flx-grow, so the span is full width and the tooltip is displayed in the middle

galloween: actually, meaybe there should be 2 different spans - so that the tooltip directive is not inialized unnecessary...

<span *ngIf="!item.tooltip">{{item.label}}</span>
<span *ngIf="item.tooltip" class="flx-grow" [matTooltip]="item.tooltip">{{item.label}}</span>

igorCuznetov: Totally agree!

projects/ui-framework/src/lib/navigation/menu/menu.component.spec.ts
```diff
@@ -42,7 +43,7 @@ describe('MenuComponent', () => {
   beforeEach(waitForAsync(() => {
     TestBed.configureTestingModule({
       declarations: [MenuComponent],
-      imports: [CommonModule, NoopAnimationsModule, MatMenuModule],
+      imports: [CommonModule, NoopAnimationsModule, MatMenuModule, MatTooltipModule],
     })
```
galloween, line 47: actually you can:

```
MockDirective(MatTooltip)

 matTooltip = fixture.debugElement
        .queryAll(By.directive(MatTooltip))[0]
        .injector.get(MatTooltip);

 expect(matTooltip.message).toContain('Raising to the top');
```

galloween: (not sure about injector - i copied it from somewhere. maybe there is directiveInstance, similar to mockComponent's componentInstance)

galloween (commented):
more comments!

projects/ui-framework/src/lib/navigation/menu/menu.component.spec.ts
igorCuznetov: @galloween the trick with mockDirective doesn't work, so, I've removed the test, as you suggested

PR #2140, opened 2021-05-25
New contentTemplate approach & 'show all' changes

MikiAsh:


projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.html
```diff
@@ -1,4 +1,11 @@
...
+  <ng-container *ngTemplateOutlet="contentTemplate; context: { item: item, index: i }">
   </ng-container>   
 </div>
+
+<b-button class="mrg-4"
+  [type]="buttonType.tertiary"
+  [size]="buttonSize.small"
+  [text]="showAll ? ('common.view-less' | translate) : ('common.view-all' | translate)"
+  (clicked)="showAll = !showAll">
+</b-button>
```
MikiAsh, line 11: I had to move this button from the outside template in order to implement the showAll logic.
Moving it here meant doing some css changes

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss
```diff
@@ -5,11 +5,30 @@
   flex-direction: column;
 }
 
+.item-container[data-expanded='true'] {
+  max-height: 419px;
```
MikiAsh, line 9: I'm starting with a naive fixed value. I'm waiting on an answer from Rotem about the increase ratio

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.html
```diff
@@ -1,4 +1,14 @@
-<div class="list-item" *ngFor="let item of items; let i = index">
-  <ng-container *ngTemplateOutlet="contentChild?.tpl; context: { item: item, index: i }">
-  </ng-container>   
+<div class="item-container" [attr.data-expanded]="showAll ? true : null">
```
galloween, line 1: you can leave just ="showAll"

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss
galloween, line 9: how do you know the max-height? it probably should be calculated from the item height, which may be different...

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss
```diff
@@ -5,11 +5,30 @@
...
   align-items: center;
+  padding: 0 16px;
   &:not(:last-child) {
     border-bottom: 1px solid $grey-400;
   }
 }
+
+b-button ::ng-deep button {
+  color: $grey-700;
+  font-weight: 700 !important;
```
galloween, line 25: - you can add class clr-grey-700 to the b-button
- font-size 700 does not do anything, because the font we use has max weight 600, and its already 600 on the button type 3 size small

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss
```diff
@@ -5,11 +5,30 @@
...
+  height: 27px !important;
+  &:hover {
+    background-color: $grey-200;
+  }
+}
+
+ b-divider::before,  b-divider::after {
+  border-color: $grey-400 !important;
+}
```
galloween, line 34: there is no point  in using b-divider component if you override its only 2 css styles (margin and border).
just use css border.

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.html
```diff
@@ -1,4 +1,14 @@
...
+  <div class="list-item" *ngFor="let item of (showAll ? items : (items | slice:0:defaultNumOfItems)); let i = index">
```
galloween, line 2: as item interface is unknown, its a bit problematic to implement trackBy, but you should still do it, tracking with index at least:
```
*ngFor="... trackBy: '$index'|trackByProp"
```
```

File: projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss [hunk 7]
```diff
@@ -5,11 +5,30 @@
...
+
 .list-item {
   height: 70px;
   display: flex;
   align-items: center;
+  padding: 0 16px;
   &:not(:last-child) {
     border-bottom: 1px solid $grey-400;
   }
```
galloween, line 20: not sure about the height. ideally this should be calculated from the provided items...
i think maybe the **only** style added should be the border.

and by the way, grey-400 for the border (here and other places) looks like a mistake. lets use default border color (which is $border-color, which is grey-500), unless the designers make a fuss.

the border could also be a class:
```
*ngFor=" ... let last=last; ... "  
[class.brd-b]="!last"
```

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss
```diff
@@ -5,11 +5,30 @@
...
 }
+
+b-button ::ng-deep button {
+  color: $grey-700;
+  font-weight: 700 !important;
+  height: 27px !important;
```
galloween, line 26: you can avoid !important by using this selector:
`b-button ::ng-deep button.small`

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.ts
```diff
@@ -1,18 +1,19 @@
...
+export class ListLayoutComponent extends ContentTemplateConsumer {
   @Input() items: any[];
-
-  constructor() { }
+  showAll: boolean;
+  readonly buttonType = ButtonType;
+  readonly buttonSize = ButtonSize;
+  readonly defaultNumOfItems = 3;
 
   ngOnInit() { }
```
galloween, line 18: this can be removed

projects/ui-framework/src/lib/layout/widget-box/widget-box.component.html
```diff
@@ -4,13 +4,6 @@
     <ng-content select="[header-right]"></ng-content>
   </div>
 </header>
-<div class="content pad-16 brd-b">
+<div class="content pad-t-16">
   <ng-content select="[content]"></ng-content>
```
galloween, line 8: for the main content projection slot i suggest you dont use the `select` here (but you can still keep the attribute in example/notes)

galloween (commented):
some html/css comments,
main question is - the item height can't be assumed, it should be calculated from the item's passed. but this can be added in next pr, of course

projects/ui-framework/src/lib/layout/widget-box/list-layout/list-layout.component.scss
MikiAsh: I wanted to leave for later PR but... added it to this one since you commented :)

galloween (approved):
more minor comments

PR #2155, opened 2021-05-31
added the navigation buttons on the ee-layout component. 

kobrigo:
closes https://app.asana.com/0/1159939249943132/1200279780148686/f

image

description allow to disable the buttons, hide them, add a tooltip to them and supply handlers when clicking them

### DIFF
--- a/projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.html
+++ b/projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.html
@@ -12,30 +12,28 @@
        [ngStyle]="config?.sidebarStyle">
 
     <div class="bel-navigation-container" *ngIf="avatar">
-      <div class="nav-button-container">
-        <b-round-button *ngIf="showPrev"
-                        type="tertiary"
-                        [matTooltip]="tooltipPrev"
-                        [disabled]="disablePrev"
-                        (clicked)="prevClicked.emit($event)"
-                        [icon]="icons.chevron_left">
-        </b-round-button>
-      </div>
+      <b-round-button *ngIf="showPrev"
+                      class="bel-nav-button-prev"
+                      type="tertiary"
+                      [matTooltip]="tooltipPrev"
+                      [disabled]="disablePrev"
+                      (clicked)="prevClicked.emit($event)"
+                      [icon]="icons.chevron_left">
+      </b-round-button>
 
       <b-avatar class="bel-avatar"
                 [avatar]="avatar"
                 [size]="avatarSize.medium"
                 [orientation]="avatarOrientation.vertical"></b-avatar>
 
-      <div class="nav-button-container">
-        <b-round-button *ngIf="showNext"
-                        type="tertiary"
-                        [matTooltip]="tooltipNext"
-                        [disabled]="disableNext"
-                        (clicked)="nextClicked.emit($event)"
-                        [icon]="icons.chevron_right">
-        </b-round-button>
-      </div>
+      <b-round-button *ngIf="showNext"
+                      class="bel-nav-button-next"
+                      type="tertiary"
+                      [matTooltip]="tooltipNext"
+                      [disabled]="disableNext"
+                      (clicked)="nextClicked.emit($event)"
+                      [icon]="icons.chevron_right">
+      </b-round-button>
     </div>
 
     <ng-content select="[ee-layout-sidebar]"></ng-content>
--- a/projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.html
+++ b/projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.html
@@ -11,19 +11,19 @@
        [ngClass]="config?.sidebarClass"
        [ngStyle]="config?.sidebarStyle">
 
-    <div class="bel-navigation-container">
-      <div class="nav-button-container" >
+    <div class="bel-navigation-container" *ngIf="avatar">
+      <div class="nav-button-container">
         <b-round-button *ngIf="showPrev"
                         type="tertiary"
-                        [matTooltip]="config.prevTooltip"
+                        [matTooltip]="tooltipPrev"
+                        [matTooltipHideDelay]="10000000"
                         [disabled]="disablePrev"
                         (clicked)="prevClicked.emit($event)"
                         [icon]="icons.chevron_left">
         </b-round-button>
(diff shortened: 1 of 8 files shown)


projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.scss
```diff
@@ -45,6 +70,14 @@ $ee-layout-grid-gap: $b-grid-gap;
   grid-template-columns: $ee-layout-sidebar-width;
 }
 
+:host {
+  &[data-wide-sidebar=true] {
+    .bel-sidebar {
+      grid-template-columns: $ee-layout-wide-sidebar-width;
```
galloween, line 76: you also need to set the width to same var

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.interface.ts
```diff
@@ -19,4 +19,9 @@ export interface EELayoutConfig {
 
   contentFooterClass?: string | string[] | NgClass;
   contentFooterStyle?: GenericObject<string>;
+
+  nextTooltip?: string;
+  prevTooltip?: string;
```
galloween, line 24: i guess we need to make them inputs as well, because tooltips can also be dynamic.

so its 6 properties already.. maybe lets make it one [navButtonsConfig] input  of type EELayoutNavButtonsConfig ?
will work just as good as separate inputs.

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.scss
galloween: another approach is to set everything to use a css variable:
```
 width: var(--bel-sidebar-width, #{$ee-layout-sidebar-width});
 grid-template-columns: var(--bel-sidebar-width, #{$ee-layout-sidebar-width});
```
the weird #{$ notation in the var() is to give the variable fallback value

and then set the variable value on host:
```
:host[data-wide-sidebar='false'] {
   --bel-sidebar-width: 180px;
}
:host[data-wide-sidebar='true'] {
   --bel-sidebar-width: 240px;
}
```

scss variables are no longer needed.

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.html
```diff
@@ -11,11 +11,33 @@
...
 
-    <b-avatar *ngIf="avatar"
-              class="bel-avatar"
-              [avatar]="avatar"
-              [size]="avatarSize.medium"
-              [orientation]="avatarOrientation.vertical"></b-avatar>
+    <div class="bel-navigation-container">
+      <div class="nav-button-container" >
+        <b-round-button *ngIf="showPrev"
+                        type="tertiary"
```
galloween, line 17: - there is no need for nav-button-container element
- the bel-navigation-container will exist even if there is no buttons or avatar inside it

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.stories.ts
```diff
@@ -66,8 +67,13 @@ const template = `
 <b-ee-layout
       [type]="type === 'null' ? null : type"
       [config]="config"
+      [showNext]='showNext'
+      [showPrev]='showPrev'
+      [disableNext]='disableNext'
+      [disablePrev]='disablePrev'
```
galloween, line 73: should be double quotes around attribute values

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.scss
```diff
@@ -12,13 +13,37 @@ $ee-layout-grid-gap: $b-grid-gap;
...
+  justify-content: space-between;
+  align-items: flex-start;
+
+  .nav-button-container {
+    min-width: $button-size-medium;
+  }
+
+  b-round-button {
+    position: relative;
+    top: 27px;
```
galloween, line 38: let's leave a comment for future generations:
```
// avatar height is 90, button height is 36 
// offset to position button in the middle of the avatar: 90/2-36/2=27
```

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.html
galloween: i guess the buttons wont make any sense without the avatar inside, so how about moving the avatar's ngIf to the container ?
```
 <div *ngIf="avatar" class="bel-avatar-container">
     <b-round-button *ngIf="navButtonsConfig?.showPrev" ...>
     <b-avatar. ...>
     <b-round-button *ngIf="navButtonsConfig?.showNext"  ...>
 </div>
```

galloween (commented):
see comments

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.stories.ts
kobrigo: yes good catch, I missed this. Just changed the webstorm default completion of the quotes to be double

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.scss
kobrigo: 👍 . I've looked for the constants to do the calac in the code but it was too complicated since some of them are only in TypeScript. cool I'll do that.

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.component.scss
kobrigo: cool, I'll do that. I did not get to work with css variables yet. niiiice one.

projects/ui-framework/src/lib/layout/ee-layout/ee-layout.interface.ts
kobrigo: If you ask me, I prefer the inputs for dynamic interactions. one of the benefit is that it's easier to play with on the storybook UI test it (easier than editing a json). 
if it's a config the developer has to think about immutability when changing the config structure right?

but whatever you think.

kobrigo: I decided to go with inputs.

galloween (approved):

PR #2158, opened 2021-06-01
Add info-graphic-text component and module

MaxDsr:
develop new component related to the new design:

https://app.asana.com/0/1200306097679416/1200306097679472/f

Screenshot 2021-06-01 at 14 33 55

### DIFF
--- a/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
+++ b/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
@@ -1,0 +1,10 @@
+<b-display-2 class="text-center mrg-b-16">{{title}}</b-display-2>
+<div class="items-container">
+  <b-display-4 class="mrg-r-16" *ngFor="let item of items" [innerHTML]="item"></b-display-4>
+</div>
+
+<div class="flx flx-row-align-x mrg-t-24">
+  <b-text-button [button]="buttonConfig">
+    <b-icon [config]="linkIconConfig"></b-icon>
+  </b-text-button>
+</div>
--- a/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
+++ b/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
@@ -1,6 +1,6 @@
-<b-display-2 class="text-center mrg-b-16">{{title}}</b-display-2>
+<div class="b-display-2 text-center mrg-b-16">{{title}}</div>
 <div class="items-container">
-  <b-display-4 class="mrg-r-16" *ngFor="let item of items" [innerHTML]="item"></b-display-4>
+  <span class="b-display-4 mrg-x-8" *ngFor="let item of items" [innerHTML]="item"></span>
 </div>
 
 <div class="flx flx-row-align-x mrg-t-24">
--- a/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
+++ b/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
@@ -1,10 +1,8 @@
-<div class="b-display-2 text-center mrg-b-16">{{title}}</div>
-<div class="items-container">
-  <span class="b-display-4 mrg-x-8" *ngFor="let item of items" [innerHTML]="item"></span>
+<div class="b-display-2 clr-negative-500 text-center mrg-b-16">{{title}}</div>
+<div class="flx flx-wrap">
+  <span class="b-display-4 clr-negative-500 mrg-x-8" *ngFor="let item of items" [innerHTML]="item"></span>
 </div>
 
-<div class="flx flx-row-align-x mrg-t-24">
-  <b-text-button [button]="buttonConfig">
+<b-text-button class="text-center mrg-t-24" [button]="buttonConfig">
     <b-icon [config]="linkIconConfig"></b-icon>
-  </b-text-button>
-</div>
+</b-text-button>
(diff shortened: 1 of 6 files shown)


projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
```diff
@@ -0,0 +1,10 @@
+<b-display-2 class="text-center mrg-b-16">{{title}}</b-display-2>
+<div class="items-container">
+  <b-display-4 class="mrg-r-16" *ngFor="let item of items" [innerHTML]="item"></b-display-4>
```
galloween, line 3: all the typography components do is just add css.
we have them historically, but its preferable just to use a class name.
even such a simple component will still add unnecessary overhead that is easily avoided by using
`<span class="b-display-4 ... "` instead

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
```diff
@@ -0,0 +1,10 @@
...
+</div>
+
+<div class="flx flx-row-align-x mrg-t-24">
+  <b-text-button [button]="buttonConfig">
+    <b-icon [config]="linkIconConfig"></b-icon>
+  </b-text-button>
```
galloween, line 9: i see what you did there. 
i added some styles (rebase), so you now can add 'b-icon-large' class to the b-text-button 
and it will modify the icon inside (same with other sizes, of course, and also colors).
the icon type you can put in the buttonConfig

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  @Input() buttonConfig: ButtonConfig = {
+    text: '',
+    type: ButtonType.negative,
+    onClick: () => console.log('button clicked !')
+  };
+  @Input() linkIconConfig: Icon = {
+    size: IconSize.large,
+    color: IconColor.negative,
+    icon: Icons.chevron_right
+  };
```
galloween, line 23: so, put `icon: Icons.chevron_right` in the button config, and add class `b-icon-large` on the text-button to modify the size

galloween: oh, didnt notice it was an input. i dont think it needs to be an input! for now, we only have 1 use-case for it. 
so i think the button and the icon configs can be hardcoded for now.

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
```diff
@@ -0,0 +1,23 @@
+@import '../../style/common-imports';
+
+:host {
+  display: block;
+  background-color: rgba(red($negative-500), green($negative-500), blue($negative-500), 0.05);
```
galloween, line 5: what are you doing here? :) if you need negative-500 with opacity 0.05, you can just do 
```
rgba($negative-500, 0.05)
```
sass knows what to do

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
```diff
@@ -0,0 +1,23 @@
...
+  border-radius: 8px;
+  padding: times8(4) times8(6);
```
galloween, line 7: judging by figma, should be equal `padding: 40px`

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
```diff
@@ -0,0 +1,23 @@
...
+}
+
+b-display-2, b-display-4 {
+  color: $negative-500;
+}
```
galloween, line 12: try `<span class="b-display-2 clr-negative-500">` (clr class should come after the display, to override the color)

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
```diff
@@ -0,0 +1,10 @@
+<b-display-2 class="text-center mrg-b-16">{{title}}</b-display-2>
+<div class="items-container">
+  <b-display-4 class="mrg-r-16" *ngFor="let item of items" [innerHTML]="item"></b-display-4>
+</div>
+
+<div class="flx flx-row-align-x mrg-t-24">
```
galloween, line 6: make :host { display: block; justify-content: center; } and then you dont need this wrapper (move the margin to the button) or the text-center classes

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
galloween: instead of mrg-r-16, use mrg-x-8, then they will align better

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.items-container {
+  display: flex;
+  flex-wrap: wrap;
+}
```
galloween, line 17: `<div class="flx flx-wrap"`

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
MaxDsr: @galloween sounds nice, but by design the icon is placed after the text. If I'll do like you suggest, it will be placed before the text.

PR #2167, opened 2021-06-03
Select and view - add truncate tooltip

rinat-shemer:
Screen Shot 2021-06-03 at 16 44 04

### DIFF
--- a/projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
+++ b/projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
@@ -22,7 +22,7 @@
 }
 
 .viewItem {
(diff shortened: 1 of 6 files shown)


projects/ui-framework/src/lib/lists/select-and-view/select-and-view.component.scss
```diff
@@ -22,7 +22,7 @@
 }
 
 .viewItem {
-  grid-template-columns: 40px auto 40px;
+  grid-template-columns: 40px calc(100% - 80px) 40px;
```
galloween, line 25: to make something grow, you can use `1fr`: 40px 1fr 40px (1fr is grid's version of flex-grow:1)

galloween (approved):

PR #2182, opened 2021-06-07
Style component corresponding to new design

MaxDsr:
image

### DIFF
--- a/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
+++ b/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
@@ -8,6 +8,7 @@
   </span>
 </div>
 
-<b-text-button class="text-center mrg-t-24" [button]="buttonConfig">
+<b-text-button class="text-center mrg-t-24 b-display-3" [button]="buttonConfig">
+    <span class="button-text b-display-4">{{buttonText}}</span>
     <b-icon [config]="linkIconConfig"></b-icon>
 </b-text-button>
--- a/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
+++ b/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
@@ -8,7 +8,7 @@
   </span>
 </div>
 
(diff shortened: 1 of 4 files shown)


projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.html
```diff
@@ -8,6 +8,7 @@
   </span>
 </div>
 
-<b-text-button class="text-center mrg-t-24" [button]="buttonConfig">
+<b-text-button class="text-center mrg-t-24 b-display-3" [button]="buttonConfig">
+    <span class="button-text b-display-4">{{buttonText}}</span>
     <b-icon [config]="linkIconConfig"></b-icon>
```
galloween, line 13: so which one you want -  b-display-3 or 4 ? 

image

galloween: you can also bind text as input: `[text]="buttonText"`
it works fine together with [button] input. 
usually  button or icon or similar config inputs are for static properties, and dynamic properties are bound via separate inputs.

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.component.ts
```diff
@@ -15,9 +15,7 @@ export class InfoGraphicTextComponent {
 
   @Input() items: ColorTextItem[];
   @Input() title: string;
-  @Input() set buttonText(value: string) {
-    this.buttonConfig.text = value || '';
-  }
```
galloween, line 20: didnt work because object had same reference.
could be fixed with this.buttonConfig = { ...this.buttonConfig, text }, 
but separate input on the button is better anyway!

MaxDsr: @galloween it worked fine, the idea is that I need to have more custom styles for the text inside this button element. That's why I place it as a separate element and define the given text there.

galloween: Ha it worked only because your component is not onPush.

galloween:
I also wonder wouldn't it be better if the bullets are before the item and not after, like in a list...

PR #2183, opened 2021-06-07
Horizontal Layout - Use new @Inputs

MikiAsh:
@Input() cardWidth: number;
  @Input() swiper: boolean = false;

### DIFF
--- a/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
+++ b/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
@@ -11,21 +11,30 @@ import { Subject } from 'rxjs';
   selector: 'b-horizontal-layout',
   templateUrl: 'horizontal-layout.component.html',
   styleUrls: ['./horizontal-layout.component.scss'],
-  changeDetection: ChangeDetectionStrategy.OnPush
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
-
-export class HorizontalLayoutComponent extends ContentTemplateConsumer implements OnDestroy {
-  @Input() items: any[];
+export class HorizontalLayoutComponent
+  extends ContentTemplateConsumer
+  implements OnDestroy {
   showAll: boolean;
+  itemsPerRow: number;
+
+  @Input() items: any[];
+  @Input() cardWidth: number;
+  @Input() swiper: boolean = false;
+  @ViewChild(CardsLayoutComponent, { static: true })
+  private cardsLayout: CardsLayoutComponent;
+
   readonly buttonType = ButtonType;
   readonly buttonSize = ButtonSize;
-  itemsPerRow: number;
   readonly cardType: CardType.small;
   private onDestroyNotifier$ = new Subject();
-  
-  @ViewChild(CardsLayoutComponent, { static: true }) private cardsLayout: CardsLayoutComponent;
 
-  constructor(private elRef: ElementRef, private DOM: DOMhelpers, private cdr: ChangeDetectorRef) {
+  constructor(
+    private elRef: ElementRef,
+    private DOM: DOMhelpers,
+    private cdr: ChangeDetectorRef
+  ) {
     super();
   }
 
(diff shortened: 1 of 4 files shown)


projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
```diff
@@ -11,21 +11,30 @@ import { Subject } from 'rxjs';
...
-  @ViewChild(CardsLayoutComponent, { static: true }) private cardsLayout: CardsLayoutComponent;
 
-  constructor(private elRef: ElementRef, private DOM: DOMhelpers, private cdr: ChangeDetectorRef) {
+  constructor(
+    private elRef: ElementRef,
+    private DOM: DOMhelpers,
+    private cdr: ChangeDetectorRef
+  ) {
     super();
   }
```
galloween, line 39: why do people put constructor in the middle of the code?
why not put it on top - after all, it's more like meta-data, inst it?

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
```diff
@@ -39,23 +48,23 @@ export class HorizontalLayoutComponent extends ContentTemplateConsumer implement
...
+      .getCardsInRow$()
+      .pipe(takeUntil(this.onDestroyNotifier$))
+      .subscribe((data) => {
+        this.itemsPerRow = data;
+        this.cdr.detectChanges();
 
-      this.DOM.setCssProps(this.cardsLayout.hostElRef.nativeElement, {
-        '--item-height': `${itemHeight}px`,
-        '--container-max-height': `calc((${itemHeight}px + var(--item-grid-gap)) * 2)`
+        const itemHeight = this.cardsLayout.hostElRef.nativeElement.querySelector('div.cards-list > div').offsetHeight;
```
galloween, line 58: children of cards-list are not guaranteed to be div's. but what **is** guaranteed is that they will be children if cards-list.

`this.cardsLayout.cardsList.children[0]`

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
```diff
@@ -39,23 +48,23 @@ export class HorizontalLayoutComponent extends ContentTemplateConsumer implement
...
+        this.DOM.setCssProps(this.cardsLayout.hostElRef.nativeElement, {
+          '--item-height': `${itemHeight}px`,
+          '--container-max-height': `calc((${itemHeight}px + var(--item-grid-gap)) * 2)`,
```
galloween, line 61: this value never changes:
`calc((var(--item-height) + var(--item-grid-gap)) * 2)` so you can just put it in css.

 though im not sure how  1 time the height and 2 times the gap are max-height. 
max height of an item with known height equals the height, no?

galloween: (the gap is between the items, not on the sides)

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
MikiAsh: Not according to the official Angular style-guide and code samples. They advise putting all class members first including @inputs and @outputs, public first and then private and after all that constructor and then life-cycle events and methods again public first. Makes sense to me, first props and then things that have () at the end (ctor/methods...)

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
MikiAsh: I moved the calc to the css. About the gap it also determines the vertical gap between lines.
I made a small change in the calc but basically it is the container's max height and it's design to kick in when when you click 'show all' and provide scroll so I'm letting it to be 2 row height + one row gap

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
galloween: 1) angular docs are notoriously outdated. what you see is probably 7 years old.
2) all javascript world except angular's examples puts constructor first. because its "meta-data"

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
```diff
@@ -39,23 +48,23 @@ export class HorizontalLayoutComponent extends ContentTemplateConsumer implement
...
-        '--container-max-height': `calc((${itemHeight}px + var(--item-grid-gap)) * 2)`
+    this.cardsLayout
+      .getCardsInRow$()
+      .pipe(takeUntil(this.onDestroyNotifier$))
+      .subscribe((data) => {
+        this.itemsPerRow = data;
+        this.cdr.detectChanges();
+        
+        const cardsList = this.cardsLayout.hostElRef.nativeElement.firstElementChild;
+        const itemHeight = cardsList.firstElementChild.offsetHeight;
```
galloween, line 59: cardsLayout has cardsList as a prop, maybe just make it public. but this is also fine

galloween (approved):

PR #2195, opened 2021-06-09
Pager for API usage

maximblack:
Fixing 2 problems:
1. If using for the API, `pageChange` & `sliceSizeChange` events are doubling API requests)
2. Added optional `resetToFirstPage` config param (when changing the slice size, navigate to the first page)

### DIFF
--- a/projects/ui-framework/src/lib/navigation/pager/pager.component.ts
+++ b/projects/ui-framework/src/lib/navigation/pager/pager.component.ts
@@ -84,7 +84,7 @@ export class PagerComponent<T = any> implements OnInit {
   @Output() sliceChange: EventEmitter<number[] | T[]> = new EventEmitter();
   @Output() pageChange: EventEmitter<number> = new EventEmitter();
   @Output() sliceSizeChange: EventEmitter<number> = new EventEmitter();
-  @Output() stateChange: EventEmitter<PagerState> = new EventEmitter();
+  @Output() stateChange: EventEmitter<PagerState<T>> = new EventEmitter();
 
   public totalItems: number;
   public totalPages: number;
@@ -102,8 +102,7 @@ export class PagerComponent<T = any> implements OnInit {
       this.initSliceConfigAndOptions();
     }
     if (!this.pagesViewModel) {
-      this.initViewModel();
-      this.emitChange();
+      this.resetState();
     }
   }
 
@@ -140,12 +139,17 @@ export class PagerComponent<T = any> implements OnInit {
     this.totalPages = Math.ceil(this.totalItems / this.config.sliceSize);
 
     const newPage =
-      this.config.resetToFirstPage === true
+      this.config.resetOnSliceSizeChange === true
         ? 0
         : Math.floor(currentSliceStart / this.config.sliceSize);
 
     this.changePage(newPage, false);
     this.emitChange('slicesize');
+  }
+
+  public resetState() {
+    this.initViewModel();
+    this.emitChange();
   }
 
   private changePage(newPage: number, emit = true): void {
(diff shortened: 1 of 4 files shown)


projects/ui-framework/src/lib/navigation/pager/pager.component.ts
```diff
@@ -138,9 +139,12 @@ export class PagerComponent<T = any> implements OnInit {
...
     this.totalPages = Math.ceil(this.totalItems / this.config.sliceSize);
 
-    const newPage = Math.floor(currentSliceStart / this.config.sliceSize);
+    const newPage =
+      this.config.resetToFirstPage === true
+        ? 0
+        : Math.floor(currentSliceStart / this.config.sliceSize);
 
-    this.changePage(newPage);
+    this.changePage(newPage, false);
```
galloween, line 147: you will still get 1 pageChange emit and 1 sliceChange emit.
but this change makes the emitChange method more confusing...

projects/ui-framework/src/lib/navigation/pager/pager.component.ts
```diff
@@ -210,8 +214,15 @@ export class PagerComponent<T = any> implements OnInit {
...
     if (which === 'slicesize') {
+      this.pageChange.emit(this.currentPage);
       this.sliceSizeChange.emit(this.config.sliceSize);
     }
+
+    this.stateChange.emit({
+      page: this.currentPage,
+      limit: this.config.sliceSize,
+      offset: this.currentPage * this.config.sliceSize
+    });
```
galloween, line 225: the state should be currentPage, sliceSize, currentSlice (number[])

galloween (commented):
see comments, mainly about PagerState properties.

also, if your implementation calls API 2 times, its not the pager's fault - its the implementation's fault!

there are multiple ways to fix it: 
- use `onlyDistinct` operator before the api call - so its not called 2 times with same request, but will call again if you go back to already visited page
- you can even use `cacheSwitchMap` operator so the api will be called only 1 time for any unique request/pager state

having pagerState output is also ok, if it will work better for your implementation

maximblack:
> see comments, mainly about PagerState properties.
> 
> also, if your implementation calls API 2 times, its not the pager's fault - its the implementation's fault!
> 
> there are multiple ways to fix it:
> 
> * use `onlyDistinct` operator before the api call - so its not called 2 times with same request, but will call again if you go back to already visited page
> * you can even use `cacheSwitchMap` operator so the api will be called only 1 time for any unique request/pager state
> 
> having pagerState output is also ok, if it will work better for your implementation

what I'm trying to solve is a situation like this:
1. Change the page 1 -> 2
2. Change the slice size 50 > 25

in order to pass the values to API, I'm subscribing to `pageChange` and `sliceSizeChange` events
in the old implementation, it will trigger 1 `pageChange` event  1 `sliceSizeChange` event, which are different
I can't use `onlyDistinct` or `cacheSwitchMap` here, only `debounceTime` can help

projects/ui-framework/src/lib/navigation/pager/pager.component.ts
maximblack: `currentSlice` is not usefull for API requests, cause we are using either `[limit, offset]` or `[currentPage, limit]`

galloween:
debounceTime is also great :)
but i see your point.

ok then instead of adding pageChange emit to `which === 'slicesize'`, 
add `|| which === 'slicesize'` to `(which === 'page' || which === 'all')`
hope it makes sense :)
 i want to keep only 1 .emit() call of each event type in the emitChange method. but you can add more rules to the IFs

projects/ui-framework/src/lib/navigation/pager/pager.component.ts
galloween: ok, false is fine, but please modify the emitChange method to have only 1 .emit() call of each even type (use the IFs for conditions of each call)

PR #2199, opened 2021-06-10
fixed show more issue and rm class

SkyWallkeRed:
rm class instead of prev pr, added logic to handle hiding "show more" if data lengt is 1 or less
connected story controller to ui

### DIFF
--- a/projects/ui-framework/src/lib/layout/insights-panel/insights-panel.component.html
+++ b/projects/ui-framework/src/lib/layout/insights-panel/insights-panel.component.html
@@ -53,7 +53,7 @@
       <ng-container *ngTemplateOutlet="item; context: {
                         insight: insight,
                         last: !isExpanded && index+1 === config.showMoreAfterItem,
-                        hasShowMore: !isExpanded && index+1 === config.showMoreAfterItem && data.length > 1,
+                        hasShowMore: !isExpanded && index+1 === config.showMoreAfterItem && data.length > config.showMoreAfterItem,
                         hasShowLess: false
                       }">
       </ng-container>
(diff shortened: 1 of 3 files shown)


projects/ui-framework/src/lib/layout/insights-panel/insights-panel.component.html
```diff
@@ -53,8 +53,7 @@
       <ng-container *ngTemplateOutlet="item; context: {
                         insight: insight,
                         last: !isExpanded && index+1 === config.showMoreAfterItem,
-                        hasShowMore: !isExpanded && index+1 ===
-                        config.showMoreAfterItem,
+                        hasShowMore: !isExpanded && index+1 === config.showMoreAfterItem && data.length > 1,
```
galloween, line 56: should it be   `data.length > config.showMoreAfterItem` ?

projects/ui-framework/src/lib/layout/insights-panel/insights-panel.stories.ts
```diff
@@ -142,8 +142,7 @@ story.add(
         showMoreAfterItem: number('showMoreAfterItem', 1, {}, 'Props'),
         expandButtonText: text('expandButtonText', 'Insights', 'Props'),
         readMoreLinkText: text('readMoreLinkText', 'Read More', 'Props'),
-        data: data,
-        dataNotes: object('data', data, 'Data'),
+        data: object('data', data, 'Data'),
```
galloween, line 145: if you look closely at the panel now, you will understand why data was separated from the dataNotes

galloween (approved):

PR #2200, opened 2021-06-10
add new shoutout comment design and type + adding emoji in comment functionality 

talitamar91:


projects/ui-framework/src/lib/comments/edit-comment/edit-comment.component.ts
```diff
@@ -135,6 +142,34 @@ export class EditCommentComponent
...
+    textarea.style.height = 'auto';
+    this.renderer.addClass(textarea, 'textarea-nowrap');
+    const isLongLineString = textarea.scrollWidth > textarea.offsetWidth;
+    this.renderer.removeClass(textarea, 'textarea-nowrap');
+    if (textarea.value.split(/\r*\n/).length > 1 || isLongLineString) {
+      textarea.style.height = textarea.scrollHeight + 'px';      
+      return;
+    }
+    textarea.style.height = '40px'
+  }
```
galloween, line 161: can you explain what you are trying to do here? is there a design for this?
shouldn't the input be same height all the time?

projects/ui-framework/src/lib/comments/edit-comment/edit-comment.component.ts
```diff
@@ -165,6 +200,10 @@ export class EditCommentComponent
   }
 
   onBlur(): void {
+    if (!this.isHtml) {
+      this.lastCaretInputCommentPosition = this.kbrdCntrlSrvc.
+      getInputCursorState(this.commentInput.nativeElement as HTMLTextAreaElement).selectionStart;
```
galloween, line 205: i suggest using similar name, as in service - `lastInputCursorState` - and save the whole object. you will need other data from there very soon.

galloween (commented):
see comments

projects/ui-framework/src/lib/comments/edit-comment/edit-comment.component.ts
talitamar91: Actually there no design for it yet.. just for the input but without long string comment. I Spoke with Rotem (the disgner) about that so meanwhile i was do it the way i think and we can/probably change it of course. i thought to do it like in facebook comment.. because the border-radios and the scroll so just wanted the height will fit the content

projects/ui-framework/src/lib/comments/edit-comment/edit-comment.component.scss
```diff
@@ -20,7 +56,8 @@
   }
```
galloween, line 56: you can make make css more maintainable if you:

- separate the type=primary/secondary - **do not override**, just wrap relevant styles in the type=xx  block
- move all padding, borders etc from textarea to the wrapper
- do not nest classes unless necessary,  flat structure is much better

projects/ui-framework/src/lib/comments/edit-comment/edit-comment.stories.ts
```diff
@@ -84,11 +87,11 @@ story.add(
         content: text('content', 'First comment!', 'Props'),
         name: text('name', name, 'Props'),
         avatar: text('avatar', avatar, 'Props'),
-
+        type: select('type', Types, Types.primary, 'Props'),
```
galloween, line 90: `Object.values(Types).slice(0,2)` - to have just first two in the knobs

galloween (approved):
lets streamline css a little (remove all overrides - 'unset' etc) and simplify the height logic

projects/ui-framework/src/lib/comments/edit-comment/edit-comment.component.ts
talitamar91: the problem here is that if we remove the height then the min scrollHeight is 58px. its because the line-height is 1.5

projects/ui-framework/src/lib/comments/edit-comment/edit-comment.component.ts
```diff
@@ -135,6 +142,34 @@ export class EditCommentComponent
     this.inputValue = this.value;
   }
 
+  ngAfterContentChecked() {
+    if (this.type !== Types.secondary || !this.commentInput) { return; }
+    this.setTextAreaHeight();
+  }
```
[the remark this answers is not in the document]
talitamar91: the onInputChange not occurs when you break a line manually

PR #2209, opened 2021-06-14
Create a dedicated slot for last-item in horizontal-layout

MikiAsh:
and remove scrollbar and expand button when swiper=true

### DIFF
--- a/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.html
+++ b/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.html
@@ -11,9 +11,12 @@
     <ng-container *ngTemplateOutlet="contentTemplate; context: { item: item, index: i }">
     </ng-container>
   </div>
+  <div #lastItem>
+    <ng-content select="[lastItem]"></ng-content>
+  </div>
 </b-cards>
 
-<b-widget-box-expand *ngIf="items?.length > itemsPerRow"
+<b-widget-box-expand *ngIf="!swiper && items?.length > itemsPerRow"
   class="flx-col brd-t pad-4" 
   (clicked)="toggleShowAll($event)">
 </b-widget-box-expand>
--- a/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
+++ b/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
@@ -53,6 +53,7 @@ export class HorizontalLayoutComponent
 
   ngAfterContentInit(): void {
     this.setItemsPerRow();
+    !this.hasLastItem && this.lastItem.nativeElement.remove();
   }
 
   toggleShowAll(showAll: boolean): void {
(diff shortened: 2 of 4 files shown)


projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.html
```diff
@@ -11,9 +11,12 @@
     <ng-container *ngTemplateOutlet="contentTemplate; context: { item: item, index: i }">
     </ng-container>
   </div>
+  <div #lastItem>
+    <ng-content select="[lastItem]"></ng-content>
+  </div>
```
galloween, line 16: empty div will be part of cards layout. you need to remove the div if there is no lastitem

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
```diff
@@ -50,19 +53,24 @@ export class HorizontalLayoutComponent
...
 
   toggleShowAll(showAll: boolean): void {
     this.showAll = showAll;
     this.cdr.detectChanges();
   }
 
+  private get hasLastItem(): boolean {
+    return !!this.lastItem.nativeElement.childNodes.length;
+  }
+
```
galloween, line 67: this will probably return true most of the time, because `childNodes` will include all **nodes** - comment nodes as well, which angular uses a lot.

safest property to use to count children is `childElementCount`

(all elements are nodes but most nodes are not elements)

this will ignore ng-content that is just text, but maybe thats what we want.

-----------

the safest way to check if element is empty is this method: `this.DOM.isEmpty(element)` - this will consider elements and text nodes, but will ignore comments etc.

-----------

this is the approach i used to use in similar circumstances - not 100% sure it is still valid 2 years later, but the advantage is that it allows to show the element again if needed:
```
<div #someElem *ngIf="hasElem" >

  @ViewChild('someElem') prefix: ElementRef;
  hasElem = true;

  ngAfterViewInit(): void {
    this.zone.runOutsideAngular(() => {
      setTimeout(() => {
        this.hasElem = !this.DOM.isEmpty(this.someElem.nativeElement);
        if (!this.cd['destroyed']) {
          this.cd.detectChanges();
        }
      }, 0);
    });
  }
```

why like this? it starts with ngIf true - so that viewChild can get the element.
`ngAfterViewInit` hook happens last (after content init). 
the `settimeout` is to get rid of 'something changed after view was checked' error, i think, but dont remember actually. 
`runOutsideAngular` is so that timeout does not trigger global change detection

galloween:
you can merge and proceed if you feel i'm slowing you down with comments.

projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
MikiAsh: Using your code sample, this.someElem always returned undefined

galloween: this approach is currently in use in at least 9 components. so i will assume you modified the sample in some way.

PR #2217, opened 2021-06-16
css tweaks in infographic component

MikiAsh:
https://app.asana.com/0/1200306097679416/1200477720699436/f

### DIFF
--- a/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
+++ b/projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
@@ -2,31 +2,42 @@
 
 :host {
   display: flex;
-  flex-direction: column;
+  flex-direction: column; 
   align-items: center;
   background-color: rgba($negative-500, 0.05);
   border-radius: 8px;
-  padding: times8(5) times8(4);
+  padding: 40px 28px 32px;
 }
 
-.flx.flx-wrap .b-display-4 {
-  position: relative;
-
-  &.clickable:hover {
-    text-decoration: underline;
-    cursor: pointer;
-  }
-
-  &::after {
-    content: '\2219';
-    position: absolute;
-    right: -10px;
-  }
+.header {
+  height: 49px;
 }
 
-.button-text {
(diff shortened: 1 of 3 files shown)


projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
```diff
@@ -2,40 +2,49 @@
 
 :host {
   display: flex;
-  flex-direction: column;
+  flex-direction: column; 
   align-items: center;
   background-color: rgba($negative-500, 0.05);
   border-radius: 8px;
-  padding: times8(5) times8(4);
+  padding: 40px 28px 32px;
```
MaxDsr, line 9: use times8(5) instead of 40px and times8(4) instead of 32px

MaxDsr (changes requested):

projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
galloween: there is also base8 function that converts any number to closest dividable by 8:
```
base8(45) === 48 === times8(6)
```

so i was thinking, maybe the best notation is actually
```
base8(40) 28px base8(32)
```

yes we all know that 40 and 32 is dividable by 8 and base8 will return same number,  
but here the base8 just marks the fact that these numbers conform to the 'dividable by 8' spec (same as times8(5) and times8(4) would) and separate them from intentionally different middle 28px, 
but at the same time, base8(40) is kinda more readable - you dont need to multiply 8*5 to understand the value...

just thinking :)

MikiAsh: I find this times8() thing irrelevant. It's  a bit naive to think that all paddings/margins should be devisable by 8 as a way of keeping consistency, it doesn't work in real life and it's even weirder to write some values with it and some without it as @MaxDsr suggested. As for the new (for me) base8 thing... why? It seems overcomplicating things and making it unreadable.

PR #2222, opened 2021-06-17
Fix b-text-button color in infographic component

MikiAsh:


projects/ui-framework/src/lib/layout/info-graphic-text/info-graphic-text.scss
```diff
@@ -50,4 +50,8 @@ b-icon {
 b-text-button {
   flex-shrink: 0;
   padding-top: times8(2);
+  color: var(--negative-500) !important;
+  ::ng-deep b-icon > .b-icon::before {
+    color: var(--negative-500);
+  }
 }
```
galloween, line 57: mmm, why do you have to do it, and passing negative as color to text button and icon is not enough?

galloween: i see, you modify negative 600 to 500.
this should work better:
- set in css: b-text-button { --negative-600: $negative-500; }
- pass `IconColor.inherit` as linkIconConfig.color

MikiAsh: Without this it generates a different shade of negative -  --negative-600

PR #2224, opened 2021-06-17
Horizontal layout last-item slot fix

MaxDsr:
- use children in order to avoid boolean decision depending on text-nodes and comments
- add missing interface
- remove redundant type


projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.ts
```diff
@@ -62,7 +63,7 @@ export class HorizontalLayoutComponent
   }
 
   private get hasLastItem(): boolean {
-    return !!this.lastItem.nativeElement.childNodes.length;
+    return !!this.lastItem.nativeElement.children.length;
```
galloween, line 66: 👍🏻

galloween (approved):

PR #2232, opened 2021-06-25
Added skipUndefined param to InputObservable decorator

maximblack:
This will allow passing `undefined` values to the input observable decorator

My specific case is to display the preloader, can't use `null`, it has a meaningful value

### DIFF
--- a/projects/ui-framework/package.json
+++ b/projects/ui-framework/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bob-style",
-  "version": "4.3.302",
+  "version": "4.3.303",
   "peerDependencies": {
     "@angular/animations": "^11.0.5",
     "@angular/cdk": "^11.0.3",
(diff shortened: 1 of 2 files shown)


galloween:
no need for this argument

you already can pass `undefined`, and its actually default value (so you can just not pass anything).
then if you want to filter it, pass filter as second argument:

```
  @InputObservable(undefined, [filter(isDefined)])
```

galloween:
ah, actually the undefined is already ignored in the decorator.
then just use `@InputObservable()`

maximblack:
> ah, you actually WANT to get the undefined.
> then just use `@InputObservable()`
> default start value of the subject thats inside `InputObservable` is already `undefined`, so you will get it in your `|async` pipe - try it.
> 
> (the `if value != undefined` inside `InputObservable` will not allow to `next()` new undefined's, but the `undefined` start value is allowed)

that's the point, I want to next() new undefined values in time, not just the initial value

PR #2234, opened 2021-06-29
add data-type to chip-list to customize selected design at new people…

talitamar91:


galloween:
Use Tabs component instead: https://hibobio.github.io/bob-style/?path=/story/navigation--tabs

PR #2237, opened 2021-07-01
add required to quick filter bar

micaelpc:


Sagi-o (approved):

doroncy:
how does it look when its required ?

micaelpc:
@doroncy same - but with * above and not allow for de-select item (on single select)

projects/ui-framework/src/lib/search/quick-filter/quick-filter.component.html
```diff
@@ -9,6 +9,7 @@
                   [label]="quickFilterConfig.label"
                   [placeholder]="quickFilterConfig.placeholder || null"
                   [options]="quickFilterConfig.options"
+                  [required]="quickFilterConfig.required"
```
galloween, line 12: replace the `[required]="false"` below

micaelpc: DONE

doroncy (approved):

PR #2251, opened 2021-07-05
Improve var naming

MikiAsh:
and improve terminology - use vertical-scroll instead of just scroll in data attributes and methods

### DIFF
--- a/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.scss
+++ b/projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.scss
@@ -21,7 +21,7 @@
 }
 
 b-cards {
-  &[data-scroll=true] ::ng-deep .cards-list {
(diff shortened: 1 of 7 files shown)


projects/ui-framework/src/lib/layout/widget-box/horizontal-layout/horizontal-layout.component.scss
```diff
@@ -30,6 +30,10 @@ b-cards {
     padding-right: 12px;
   }
   overflow-x: hidden;
+
+  ::ng-deep .cards-list[data-swiper-mode=true] b-card.single-card {
+    width: 227px;
+  }
```
galloween, line 36: the card width is supposed to be responsive/fluid. why set it to 227 ?

MikiAsh: Mainly to achieve this effect of peaking next card
image

PR #2257, opened 2021-07-07
Angular 12 bump

harellevy:
continued from @SkyWallkeRed Great PR!
fix some stuff in css,  aot


harellevy:
added cdk + material 12

harellevy:
still need to fix date picker issues with material and cdk

galloween (approved):

PR #2260, opened 2021-07-14
Collapsible section - move pointer to chevron icon

omerhochbob:
Closes https://app.asana.com/0/1159939249943132/1200179398223830/f
Screen Shot 2021-07-14 at 16 34 59
Can't take screenshot of pointer but trust me it works

### DIFF
--- a/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
+++ b/projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
@@ -66,12 +66,9 @@ $bcp-header-min-height: 70px;
 .bcp-header {
   min-height: $bcp-header-min-height;
   padding: times8(1) times8(4);
-
-  @include icon-size(32px);
-  &:before {
-    width: map-get($icon-sizes, large);
-    display: flex;
-    justify-content: center;
+  .toggle-icon {
+    margin-left: -6px;
+    margin-right: 16px;
   }
   @include desktop {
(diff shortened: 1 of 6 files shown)


projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
```diff
@@ -66,7 +66,10 @@ $bcp-header-min-height: 70px;
 .bcp-header {
   min-height: $bcp-header-min-height;
   padding: times8(1) times8(4);
-
+  .toggle-icon {
+    margin-left: -6px;
+    margin-right: 16px;
+  }
```
galloween, line 72: you need to remove the include icon-size and before block

projects/ui-framework/src/lib/layout/collapsible-section/collapsible-section.component.scss
```diff
@@ -204,7 +203,9 @@ $bcp-header-min-height: 70px;
 // Collapse/expand animation
 
 .bcp-header[data-collapsible='true'] {
-  cursor: pointer;
+  .toggle-icon {
+    cursor: pointer;
+  }
```
galloween, line 202: you can add it to the first .toggle-icon block, no need to scope to collapsible=true, because when its false, the icon is not there anyway

galloween (changes requested):
we moved bob-style to the mono-repo, so now you need to update there

(this repo is not active any more)

